PR #19, opened 2026-05-22
fix: stream multipart request bodies

### LINKED ISSUE #18: Multipart uploads buffer the full request body in memory
## Summary

Multipart uploads currently build the entire request body in memory before sending it. That means a large file upload can turn into a large memory spike, even though the file-handling path uses reader-backed values that look like they should stream.

I checked this on `main` at `9eb2084`.

## What I found

- Multipart bodies are encoded into a `bytes.Buffer` before the request option is created:
  https://github.com/openai/openai-cli/blob/9eb2084/pkg/cmd/flagoptions.go#L494-L510
- File arguments in multipart/octet-stream paths are opened as reader-backed `fileUpload` values:
  https://github.com/openai/openai-cli/blob/9eb2084/pkg/cmd/flagoptions.go#L154-L166
  https://github.com/openai/openai-cli/blob/9eb2084/pkg/cmd/flagoptions.go#L276-L284
- The multipart encoder then copies those readers into the multipart part with `io.Copy`, so the bytes still end up in the buffer above before the request is sent:
  https://github.com/openai/openai-cli/blob/9eb2084/internal/apiform/encoder.go#L169-L201
- `openFileUpload` opens an `*os.File` and returns a `fileUpload`, but the production multipart encoding path does not close that owned file after copying it:
  https://github.com/openai/openai-cli/blob/9eb2084/pkg/cmd/flagoptions.go#L550-L581

## Why this matters

For endpoints that accept large files, the CLI can end up holding the full encoded multipart request in memory. That can make uploads slower than necessary, increase RSS significantly, or fail in constrained environments. There is also a small file descriptor leak risk because the upload file is opened for encoding but not closed by the normal request path.

## Possible fix

A good direction would be to stream multipart encoding with `io.Pipe`, so file contents are copied into the HTTP request as the client reads the body. The writer side should close owned upload files after each file reader is copied, while preserving the existing stdin behavior for `@-` / `-` inputs.

One wrinkle: the current `bytes.Buffer` body is replayable by the Go SDK for retries. A plain streaming pipe body would not be replayable, so the fix should make that behavior explicit. A few options:

1. Disable automatic retries for streamed multipart bodies.
2. Add a replayable multipart body that can reopen owned files for each retry.
3. Spool the encoded multipart body to a temporary file when replayability is required.

The second option is probably the most complete long-term fix. The first option may be enough for a smaller CLI-side change if the retry tradeoff is acceptable.

2830500285:
Closes #18.

## Summary
- stream multipart/form-data request bodies with `io.Pipe` instead of buffering the full encoded payload in memory
- close owned `fileUpload` readers after multipart encoding completes
- disable SDK retries for streamed multipart bodies because the body cannot be safely replayed

## Validation
- `go test ./internal/apiform ./pkg/cmd -run 'TestEncode|TestMultipartRequestBodyStreamsAndClosesUploads|TestEmbedFiles|TestEmbedFilesStdin|TestEmbedFilesUploadMetadata'`
- `go test ./... -run 'TestEncode|TestMultipartRequestBodyStreamsAndClosesUploads|TestEmbedFiles|TestEmbedFilesStdin|TestEmbedFilesUploadMetadata'`
- `go build ./...`

### DIFF
--- a/pkg/cmd/flagoptions.go
+++ b/pkg/cmd/flagoptions.go
@@ -1,7 +1,6 @@
 package cmd
 
 import (
-	"bytes"
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
@@ -492,22 +491,18 @@ func flagOptions(
 	case EmptyBody:
 		break
 	case MultipartFormEncoded:
-		buf := new(bytes.Buffer)
-		writer := multipart.NewWriter(buf)
-
 		// For multipart/form-encoded, we need a map structure
 		bodyMap, ok := requestContents.Body.(map[string]any)
 		if !ok {
 			return nil, fmt.Errorf("Cannot send a non-map value to a form-encoded endpoint: %v\n", requestContents.Body)
 		}
 		encodingFormat := apiform.FormatBrackets
-		if err := apiform.MarshalWithSettings(bodyMap, writer, encodingFormat); err != nil {
-			return nil, err
-		}
-		if err := writer.Close(); err != nil {
-			return nil, err
-		}
-		options = append(options, option.WithRequestBody(writer.FormDataContentType(), buf))
+		contentType, body := newMultipartRequestBody(bodyMap, encodingFormat)
+		options = append(options,
+			option.WithRequestBody(contentType, body),
+			// Streaming request bodies cannot be replayed safely by the SDK retry loop.
+			option.WithMaxRetries(0),
+		)
 
 	case ApplicationJSON:
 		bodyBytes, err := json.Marshal(requestContents.Body)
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
Cleanup stops at the first upload whose `Close()` returns an error. With multiple multipart files, that means one failing reader prevents every later `fileUpload` from being closed, leaking the remaining owned readers on exactly the error path this PR is trying to clean up. Could we attempt all closes and return/join the errors afterwards? A two-file test with the first close failing would pin it.

markstuart-oai:
@codex review

saioai (commented):
i checked this against current main. #65 already landed the streaming upload fix and closed #18. the focused streaming, cleanup, and retry tests passed locally. the merged version also closes every upload even if one close fails. closing this as covered by #65.

sylvesterkaczmarek:
Thanks for checking this against current main. If #65 closes every upload even when one close fails, that resolves the cleanup concern I raised here.

PR #20, opened 2026-05-22
fix: make yaml formatting side-effect free

2830500285:
## Summary
- make YAML formatting return bytes like the other output formats instead of writing directly to stdout
- keep `ShowJSON` / iterator callers responsible for writing output, so YAML can be buffered, counted, and paged consistently
- add regression coverage for the formatter side effect and the `ShowJSON` YAML output path

No linked issue; I found this while looking for another runtime correctness fix after the currently open issues were already covered by existing PRs.

## Validation
- `go test ./pkg/cmd -run 'TestFormatJSON|TestShowJSONYAML|TestShowJSONIterator'`
- `go test ./... -run 'TestFormatJSON|TestShowJSONYAML|TestShowJSONIterator'`
- `go build ./...`

Note: I also attempted `go test ./...` locally, but this Windows environment is missing `/bin/bash` for the autocomplete tests and does not have the mock server running on `localhost:4010` for generated command tests.

### DIFF
--- a/pkg/cmd/cmdutil.go
+++ b/pkg/cmd/cmdutil.go
@@ -344,8 +344,7 @@ func formatJSON(res gjson.Result, opts ShowJSONOpts) ([]byte, error) {
 		if err := json2yaml.Convert(&yaml, input); err != nil {
 			return nil, err
 		}
-		_, err := opts.Stdout.Write([]byte(yaml.String()))
-		return nil, err
+		return []byte(yaml.String()), nil
 	default:
 		return nil, fmt.Errorf("Invalid format: %s, valid formats are: %s", opts.Format, strings.Join(OutputFormats, ", "))
 	}
--- a/pkg/cmd/cmdutil_test.go
+++ b/pkg/cmd/cmdutil_test.go
@@ -233,6 +233,50 @@ func TestFormatJSON(t *testing.T) {
 		require.NotContains(t, out, "\a")
 		require.Contains(t, out, `\u001b]52;c;ZGF0YQ==\u0007`)
 	})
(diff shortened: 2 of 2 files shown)


markstuart-oai:
@codex review

saioai (changes requested):
yaml output can fetch another page after reaching --max-items and fail unnecessarily. please rebase onto main, preserve its pagination limit fix, and add a regression test.

saioai:
[saibot]

requesting changes. request changes. returning YAML bytes fixes paging, but this branch can request another page after --max-items is reached and fail after printing all the requested results. resolve onto main while preserving its item-limit fixes and add a YAML regression test.

- Medium: `pkg/cmd/cmdutil.go:347` Returning YAML bytes now lets ShowJSONIterator reach its streaming loop, which calls iter.Next() before testing itemsToDisplay == 0. On real files list with --format yaml --limit 3 --max-items 3 and three long-filename files in the first page (has_more=true), an unneeded GET ?after=file-003 occurs, its 400 error… fix: Resolve onto current main, preserve its outputIterator item-limit protections, and add a YAML exact-limit/pager regression.

local checks: 1 passed; 1 head check failed; 1 base comparison failed.

limits: Linux evidence only; full mock-backed suite and native macOS/Windows not run. Unchanged-head/base September18 fixture results reused: head7/9 and base8/9; both failed runs retained honestly, with different expected failure reasons. No interactive less test.…

github checks: none reported.

reviewed commit: `490acd47efe9`. base: `9eb208407a04`.

saioai:
@2830500285

PR #21, opened 2026-05-30
fix(completion): hide internal commands and suppress fish debug logs

jeet-dekivadia:
## Summary
- filter hidden commands from shell completion when no argument has been typed yet
- discard stderr from fish's internal completion probe instead of appending it to `/tmp/fish-debug.log`
- add focused regressions for the empty-input filter and embedded fish script

## Why
The non-empty completion path already skips `child.Hidden`, but the initial empty-input branch did not. An initial tab press could therefore expose internal commands such as `__complete`, `@completion`, and `@manpages`.

The fish completion script also appended diagnostics from every tab-completion invocation to a fixed temporary file. Bash already suppresses the same internal probe's stderr. Applying the same behavior in fish avoids an unbounded cross-session debug log during normal CLI use.

## Validation
- `git diff --check`
- `go test ./internal/autocomplete -count=1`

## Remote Linux validation
- `git diff --check`
- `./scripts/bootstrap`
- `./scripts/lint`
- `./scripts/test`

### DIFF
--- a/internal/autocomplete/autocomplete.go
+++ b/internal/autocomplete/autocomplete.go
@@ -216,7 +216,9 @@ func getAllPossibleCompletions(completionStyle CompletionStyle, root *cli.Comman
 	completions := make([]ShellCompletion, 0)
 	if len(args) == 0 {
 		for _, child := range root.Commands {
-			completions = builder.createFromCommand("", child, completions)
+			if !child.Hidden {
+				completions = builder.createFromCommand("", child, completions)
+			}
 		}
 		return CompletionResult{Completions: completions, Behavior: ShellCompletionBehaviorDefault}
 	}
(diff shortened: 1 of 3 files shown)


jeet-dekivadia:
Small follow-up for review context. This patch is limited to completion behavior: hidden commands stay hidden at the shell-completion boundary, and the fish probe no longer writes debug noise to `/tmp/fish-debug.log`. The regressions cover both the hidden-command filtering and the fish script stderr behavior without touching command execution itself.

markstuart-oai:
@codex review

saioai (changes requested):
please resolve the conflicts against main and keep the remaining hidden-command fix. the fish logging change already shipped, so please preserve the newer completion code and tests.

saioai:
[saibot]

requesting changes. request changes. the hidden-command filter is useful, but this branch conflicts with main. keep that fix and preserve the newer completion code; the fish logging fix already shipped. No blocking or substantive issues found in the remaining filter.

local checks: 2 passed; 1 base comparison failed.

limits: Linux evidence only; full mock-backed suite and native macOS/Windows not run. Unchanged-head/base September18 package and 18-case protocol evidence reused; native Bash wrapper only, no native Fish/Zsh/PowerShell. Four base noargs cases fail as expected;…

github checks: none reported.

reviewed commit: `d23a9b659976`. base: `9eb208407a04`.

saioai:
@jeet-dekivadia

PR #22, opened 2026-05-30
fix(completion): stop fish from writing debug logs

jeet-dekivadia:
## Summary
- discard stderr from fish's internal completion probe instead of appending it to `/tmp/fish-debug.log`

## Why
The fish completion script currently writes diagnostics from every tab-completion invocation to a fixed temporary file. That leaves an unbounded, cross-session debug log behind during normal CLI usage and can retain command diagnostics unexpectedly. Bash already suppresses the same internal probe's stderr; fish should do the same.

## Validation
- `git diff --check`
- `go test ./internal/autocomplete -count=1`
- verified the embedded fish completion script no longer references `fish-debug`

### DIFF
--- a/internal/autocomplete/shellscripts/fish_autocomplete.fish
+++ b/internal/autocomplete/shellscripts/fish_autocomplete.fish
@@ -7,7 +7,7 @@ function ____APPNAME___fish_autocomplete
     set -l cmd $tokens[1]
     set -l args $tokens[2..-1]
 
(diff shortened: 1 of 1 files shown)


jeet-dekivadia:
Closing in favor of #21, which now includes this fish stderr fix together with focused regression coverage so reviewers have one cohesive completion-hygiene patch.

PR #31, opened 2026-07-16
fix: honor --stream true for nullable request flags

fallintoplace:
## Problem

Ten streaming commands model the request field as requestflag.Flag[*bool] so they can represent unset, true, false, and null. The command branch used cli.Command.Bool, which only recognizes a concrete bool. As a result, passing --stream true still selected the non-streaming SDK method.

The two response retrieval commands use concrete boolean flags and are unaffected.

## Fix

- add requestflag.FlagBool to read both bool and *bool values safely
- route all ten affected create, submit, edit, and generate handlers through it
- preserve nullable request serialization, including explicit null

## Coverage

- unit coverage for concrete booleans, pointer booleans, nil, and unknown flags
- end-to-end SSE coverage for every affected command shape
- explicit --stream null coverage for the non-streaming path
- a held-open SSE response that verifies the rendered event arrives before the response completes

## Validation

- ./scripts/test
- go test -race ./internal/requestflag ./pkg/cmd -run "^(TestFlagBool|TestNullableStreamFlags)$" -count=1
- go vet ./...
- ./scripts/lint

### DIFF
--- a/internal/requestflag/requestflag.go
+++ b/internal/requestflag/requestflag.go
@@ -251,6 +251,20 @@ func ExtractRequestContents(cmd *cli.Command) RequestContents {
 	return res
 }
 
+// FlagBool returns the boolean value of a request flag. Nullable boolean flags
+// expose *bool through cli.Command.Value, while cli.Command.Bool only handles
+// concrete bool values.
+func FlagBool(cmd *cli.Command, name string) bool {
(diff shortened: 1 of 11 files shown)


markstuart-oai:
@codex review

saioai (commented):
closing this as superseded by [#53]which already fixed the nullable stream flags. thanks for the fix.

PR #32, opened 2026-07-16
fix(debug): redact sensitive response headers

fallintoplace:
## Problem

Debug logging already clones and redacts request headers, but response headers were passed directly to httputil.DumpResponse. This exposed values such as Set-Cookie, response-side Authorization, and API-key headers whenever debug logging was enabled.

## Fix

- share one case-insensitive header redaction path across requests and responses
- redact direct and proxy authorization while retaining only the authentication scheme
- shallow-copy the response before logging so the response returned to callers is untouched
- keep ordinary response headers available for debugging

## Coverage

The response test covers multiple cookies, direct and proxy authorization, an API-key header, deliberately non-canonical casing, a visible request ID, and both header and response-object identity.

## Validation

- ./scripts/test
- go test -race ./internal/debugmiddleware -count=1
- go vet ./...
- ./scripts/lint

### DIFF
--- a/internal/debugmiddleware/debug_middleware.go
+++ b/internal/debugmiddleware/debug_middleware.go
@@ -17,8 +17,8 @@ type (
 
 const redactedPlaceholder = "<REDACTED>"
 
-// Headers known to contain sensitive information like an API key. Note that this exclude `Authorization`,
-// which is handled specially in `redactRequest` below.
+// Headers known to contain sensitive information like an API key. Authorization
+// headers are handled separately so their authentication scheme remains visible.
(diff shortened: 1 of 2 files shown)


markstuart-oai:
@codex review

saioai (changes requested):
requesting changes. most of this is already on main. please rebase and keep only the remaining proxy-authorization handling and its tests, preserving the newer redaction behavior.

saioai (approved):
approved. this now keeps just the remaining proxy-authorization change and preserves the existing redaction behavior.

PR #33, opened 2026-07-16
fix: keep API error output on stderr

fallintoplace:
## Summary

- write formatted API error bodies to stderr alongside the HTTP status line
- fall back to JSON for `--format-error explore`, since the interactive explorer writes to stdout
- add an end-to-end regression test that keeps stdout empty on API failures

## Why

API failures currently split their output across streams: the status line goes to stderr, while the structured response body defaults to stdout. This makes failed commands unsafe to pipe because error JSON can be mistaken for successful output.

## Test plan

- `./scripts/test`
- regression test covers exit status, empty stdout, and the status plus JSON body on stderr

### DIFF
--- a/cmd/openai/main.go
+++ b/cmd/openai/main.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"os"
 	"slices"
+	"strings"
 
 	"github.com/openai/openai-cli/pkg/cmd"
 	"github.com/openai/openai-go/v3"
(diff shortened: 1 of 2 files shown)


markstuart-oai:
@codex review

saioai (changes requested):
requesting changes for the rebase. the stderr fix is useful and passes the focused checks, but this branch conflicts with main. please resolve the conflicts while keeping the current autocomplete and header handling.

saioai (changes requested):
requesting changes. the rebase left an unused slices import, so the cli cannot build. please remove it while keeping the current autocomplete and header handling.

saioai (changes requested):
^^

saioai:
[saibot]

review pending. automatic review is blocked: Worker blocked: Review finished: request changes because an unused slices import breaks the CLI build; the identical base check passes. Evidence handoff needs coordinator reconciliation: one…. no approval yet.

github checks: none reported.

current commit: `b1edfe072a94`. base: `ee62f85cc6fe`.

saioai:
@fallintoplace

PR #37, opened 2026-07-24
fix(release): make partially published releases retryable

apcha-oai:
## Why

The v1.5.0 publish run partially uploaded its GitHub release assets, then received `404 Not Found` for two Windows uploads. GoReleaser exited before publishing the Homebrew cask, leaving `openai/homebrew-tools` on v1.4.0. A normal rerun is not reliably recoverable because the successfully uploaded assets already exist.

## Fix

Enable GoReleaser's `release.replace_existing_artifacts` so a rerun can replace previously uploaded assets, finish publishing, and reach the Homebrew cask PR. Target `next` so the setting ships with the next release; it does not retroactively change the existing v1.5.0 tag.

## Validation

- Parsed `.goreleaser.yml` and verified schema version 2 and `replace_existing_artifacts: true`.
- `git diff --check`.
- PR CI runs the full GoReleaser snapshot build.

Failed v1.5.0 run: https://github.com/openai/openai-cli/actions/runs/30107350402

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,6 +1,9 @@
 project_name: openai
 version: 2
 
+release:
+  replace_existing_artifacts: true
+
 before:
   hooks:
     - mkdir -p completions


jbeckwith-oai (commented):
Reviewed the exact head `833ba2cf46c9280da0daf087e864217bbc985174`. No substantive findings.

I verified that the pinned GoReleaser v2.15.4 supports `release.replace_existing_artifacts` and only invokes it after a duplicate-asset 422, where it deletes the exact matching release asset and retries the upload. That is the missing recovery path for the partial v1.5.0 publish: 22 assets were already present, the two failed Windows archives were absent, and GoReleaser stopped before the Homebrew publisher. The exact-head CI is green, including the full GoReleaser snapshot build, and `git diff --check` passes.

Holding the final approval stamp for Justin's explicit go-ahead.

HAYDEN-OAI (approved):
Reviewed exact head 833ba2cf46c9280da0daf087e864217bbc985174. No substantive findings. The change adds the correct GoReleaser v2.15.4 recovery path for future tagged releases: on a duplicate-asset 422, GoReleaser deletes only the same-named existing GitHub release asset and retries, allowing a rerun after partial upload to complete the release before the Homebrew cask publisher runs. I also verified the publish ordering, serialized publish concurrency, source-repo contents-write app token, tap contents and pull-requests write token, detached-tag behavior, and the documented non-retroactive scope for v1.5.0. Exact-version goreleaser check passed against this head, and current CI is green including the snapshot release build.

PR #42, opened 2026-07-30
feat: Add mutual TLS client certificate support

jbeckwith-oai:
## Summary

- add `--mtls-client-cert-file` and `--mtls-client-key-file`, with environment-variable equivalents, for API-key requests
- build a command-scoped Go HTTP transport that presents the selected certificate chain while preserving existing proxy and TLS settings
- require an explicit HTTPS base URL and prevent cross-origin redirects from presenting the client certificate
- document beta enrollment, certificate-chain behavior, safe private-key handling, and endpoint configuration

## Why

CLI users in the OpenAI Mutual TLS Beta Program need a safe file-based way to compose API-key authentication with HTTP mTLS. Go's default certificate selection can decline an explicitly configured certificate when the endpoint advertises unrelated CA hints, so the transport returns the selected pair through `GetClientCertificate` whenever the configured endpoint requests client authentication.

## Validation

- `./scripts/test` (including Windows compile check)
- `go test -race ./pkg/cmd -run 'Test(NewMTLSHTTPClient|CLIUsesMTLS)' -count=1`
- `go vet ./...`
- `./scripts/lint`
- live mTLS smoke test against `https://mtls.api.openai.com/v1/files` with the enrolled direct certificate fixture
- live mTLS smoke test with the enrolled leaf-plus-intermediate chain fixture
- thermo-nuclear and multi-axis code-quality reviews

### DIFF
--- a/pkg/cmd/mtls.go
+++ b/pkg/cmd/mtls.go
@@ -141,8 +141,7 @@ func newMTLSHTTPClient(certFile, keyFile, baseURL string) (*http.Client, error) 
 	return &http.Client{
 		Transport: transport,
 		CheckRedirect: func(request *http.Request, via []*http.Request) error {
-			if !strings.EqualFold(request.URL.Scheme, endpoint.Scheme) ||
-				!strings.EqualFold(request.URL.Host, endpoint.Host) {
+			if !sameURLOrigin(request.URL, endpoint) {
 				return http.ErrUseLastResponse
 			}
 			if len(via) >= 10 {
@@ -151,6 +150,25 @@ func newMTLSHTTPClient(certFile, keyFile, baseURL string) (*http.Client, error) 
 			return nil
 		},
 	}, nil
+}
+
+func sameURLOrigin(left, right *url.URL) bool {
+	return strings.EqualFold(left.Scheme, right.Scheme) &&
+		strings.EqualFold(left.Hostname(), right.Hostname()) &&
+		effectiveURLPort(left) == effectiveURLPort(right)
+}
+
+func effectiveURLPort(parsedURL *url.URL) string {
+	if port := parsedURL.Port(); port != "" {
+		return port
+	}
+	if strings.EqualFold(parsedURL.Scheme, "https") {
+		return "443"
+	}
+	if strings.EqualFold(parsedURL.Scheme, "http") {
+		return "80"
+	}
+	return ""
 }
 
 func readMTLSFile(kind, path string) ([]byte, error) {
--- a/pkg/cmd/mtls.go
+++ b/pkg/cmd/mtls.go
@@ -46,6 +46,10 @@ func mtlsClientFlags() []cli.Flag {
 }
 
 func configureMTLS(ctx context.Context, cmd *cli.Command) (context.Context, error) {
+	if cmd.Metadata == nil {
+		cmd.Metadata = map[string]any{}
+	}
+
 	certFile, _ := cmd.Value(mtlsClientCertFileFlag).(string)
 	keyFile, _ := cmd.Value(mtlsClientKeyFileFlag).(string)
 	baseURL := cmd.String("base-url")
--- a/pkg/cmd/mtls.go
+++ b/pkg/cmd/mtls.go
@@ -111,6 +111,19 @@ func newMTLSHTTPClient(certFile, keyFile, baseURL string) (*http.Client, error) 
 
 	transport := baseTransport.Clone()
 	transport.ResponseHeaderTimeout = openAISDKResponseHeaderTimeout
+	if proxy := transport.Proxy; proxy != nil {
+		transport.Proxy = func(request *http.Request) (*url.URL, error) {
+			proxyURL, err := proxy(request)
+			if err != nil || proxyURL == nil {
+				return proxyURL, err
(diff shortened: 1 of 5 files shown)


pkg/cmd/mtls.go
```diff
@@ -0,0 +1,153 @@
...
+	endpoint, err := url.Parse(baseURL)
+	if err != nil || !strings.EqualFold(endpoint.Scheme, "https") || endpoint.Host == "" {
+		return nil, errors.New("mTLS requires an explicit HTTPS base URL")
+	}
+
+	certPEM, err := readMTLSFile("certificate", certFile)
+	if err != nil {
+		return nil, err
+	}
+	defer clear(certPEM)
```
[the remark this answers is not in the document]
jbeckwith-oai: No code change is needed here: `clear` is a predeclared Go built-in, and this repository now targets Go 1.25 (`go 1.25.0` in `go.mod`). The branch compiles in `./scripts/lint`, the full local suite, Windows compile checks, and CI. Resolving this as a false positive.

pkg/cmd/mtls.go
```diff
@@ -0,0 +1,153 @@
...
+	if err != nil {
+		delete(cmd.Metadata, mtlsHTTPClientMetadata)
+		return ctx, err
+	}
+
+	if client == nil {
+		delete(cmd.Metadata, mtlsHTTPClientMetadata)
+	} else {
+		cmd.Metadata[mtlsHTTPClientMetadata] = client
+	}
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in `4aee172`: `configureMTLS` now initializes `cmd.Metadata` when it is nil before storing the command-scoped HTTP client. This preserves the normal urfave/cli lifecycle behavior while making direct calls defensive. Focused race tests, the full suite, Windows compile checks, `go vet`, and lint/build all pass.

HAYDEN-OAI (commented):
Found one substantive issue in HTTPS-proxy origin scoping.

pkg/cmd/mtls.go
```diff
@@ -0,0 +1,157 @@
...
+	tlsConfig := transport.TLSClientConfig
+	if tlsConfig == nil {
+		tlsConfig = &tls.Config{}
+	} else {
+		tlsConfig = tlsConfig.Clone()
+	}
+	// The user explicitly selected this pair, so present it whenever the endpoint
+	// requests client authentication instead of filtering it against the server's
+	// advertised CA names.
+	tlsConfig.GetClientCertificate = func(*tls.CertificateRequestInfo) (*tls.Certificate, error) {
```
HAYDEN-OAI, line 136: [P2] Keep the mTLS certificate off HTTPS proxy handshakes

`ProxyFromEnvironment` is preserved when this transport is cloned, and Go uses `Transport.TLSClientConfig` for the TLS handshake to an `https://` proxy before CONNECTing to the API. Because this callback unconditionally returns the configured certificate, an HTTPS proxy that requests client authentication receives the certificate and proof of possession even though it is outside `endpoint`; `CheckRedirect` never runs for that hop. Scope certificate selection to the final API hop, or reject HTTPS proxies while mTLS is enabled, so the certificate is presented only to the configured origin.

jbeckwith-oai: Fixed in `2f8462c`. The mTLS transport now rejects HTTPS proxy URLs before dialing, so the client certificate cannot be presented during the proxy TLS handshake; HTTP and SOCKS proxies remain supported. Added regression coverage and documented the limitation. Verified with the focused race test, full test harness (including Windows compile checks), `go vet ./...`, and lint/build.

pkg/cmd/mtls.go
```diff
@@ -0,0 +1,170 @@
...
+		return &certificate, nil
+	}
+	transport.TLSClientConfig = tlsConfig
+
+	return &http.Client{
+		Transport: transport,
+		CheckRedirect: func(request *http.Request, via []*http.Request) error {
+			if !strings.EqualFold(request.URL.Scheme, endpoint.Scheme) ||
+				!strings.EqualFold(request.URL.Host, endpoint.Host) {
```
HAYDEN-OAI: [P2] Normalize origin ports before deciding a redirect is cross-origin

This compares the raw `URL.Host`, so a redirect between `https://mtls.api.openai.com` and `https://mtls.api.openai.com:443` is rejected even though both identify the same origin and certificate scope. The custom client then returns the 3xx instead of following a safe redirect. Compare canonical origins (lower-cased hostname plus an explicit effective port) so this guard blocks actual cross-origin hops without rejecting an equivalent default-port spelling.

HAYDEN-OAI (approved):

pkg/cmd/mtls.go
jbeckwith-oai: Fixed in `6e99f15`. Redirect checks now compare canonical origin components: case-insensitive scheme and hostname plus an effective port (`443` for HTTPS and `80` for HTTP). This allows `https://host` ↔ `https://host:443` while continuing to reject different ports such as `:8443`. Added regression coverage for both cases. Focused race tests, the full test harness (including Windows compile checks), `go vet ./...`, and lint/build all pass.

PR #43, opened 2026-07-31
fix(output): stop pagination at max items

fallintoplace:
## Summary

- check the remaining item limit before advancing a paged iterator
- avoid fetching or consuming one extra item after `--max-items` is reached
- cover the pager path with a regression test

## Testing

- `go test ./pkg/cmd -run TestShowJSONIterator -count=1`
- `go test ./internal/...`

### DIFF
--- a/pkg/cmd/cmdutil.go
+++ b/pkg/cmd/cmdutil.go
@@ -551,10 +551,7 @@ func ShowJSONIterator[T any](iter jsonview.Iterator[T], itemsToDisplay int64, op
 		pagerOpts := opts
 		pagerOpts.Stdout = pager
 
-		for iter.Next() {
-			if itemsToDisplay == 0 {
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the complete pagination/output diff at d82a04e7b6462e2520a33773be7665c366c42efc. No blocking findings. The short-circuit condition now checks the remaining item budget before advancing the iterator, preventing an N+1 item consumption or page fetch while preserving zero, positive, and unlimited limits. Targeted tests, race testing, internal package tests, build validation, formatting, and diff checks passed locally.

markstuart-oai:
@codex review

saioai (approved):
looks good to me. checking the limit before iter.Next() prevents the extra advance, and the new test covers that case.

PR #53, opened 2026-08-11
fix(api): handle nullable stream flags

jbeckwith-oai:
## Summary

Nullable API request schemas expose `stream` as a `*bool` CLI flag value. The generated handlers previously called `cmd.Bool("stream")`, which expects a concrete boolean and can fail when the nullable flag value is nil.

This regenerates the ten affected streaming handlers with a nil-safe pointer check before selecting the streaming request path. Non-nullable stream flags remain unchanged. The generator change is in [openai/openai#1230579](https://github.com/openai/openai/pull/1230579).

Key review points:
- `pkg/cmd/*.go`: only the affected generated streaming checks changed.
- Castiron rendered 155 CLI files from the current OpenAPI/config inputs; the three-way overlay applied cleanly with zero conflicts and produced this eight-file diff.

### Automated
- `target/release/castiron generate sdk ... --language cli`: passed; 155 files rendered, formatted, and overlaid cleanly.
- `GOFLAGS=-buildvcs=false ./scripts/lint`: passed.
- `./scripts/test`: all functional Go tests passed; the script later exited nonzero when its teardown trap tried to kill an already-exited mock-server PID.
- `GOOS=windows GOARCH=amd64 go test -c ./... -o /tmp/openai-cli-windows-tests/`: passed.
- `git diff --check`: passed.

### DIFF
--- a/pkg/cmd/audiotranscription.go
+++ b/pkg/cmd/audiotranscription.go
@@ -129,7 +129,7 @@ func handleAudioTranscriptionsCreate(ctx context.Context, cmd *cli.Command) erro
 	format := cmd.Root().String("format")
 	explicitFormat := cmd.Root().IsSet("format")
 	transform := cmd.Root().String("transform")
-	if cmd.Bool("stream") {
+	if streamFlagValue, _ := cmd.Value("stream").(*bool); streamFlagValue != nil && *streamFlagValue {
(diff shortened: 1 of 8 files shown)


HAYDEN-OAI (approved):
Reviewed the nullable stream-flag handling across the generated handlers. The pointer check preserves the tri-state request semantics and only selects streaming for an explicit true value. No substantive issues found.

PR #65, opened 2026-08-15
fix: stream multipart uploads without buffering

### LINKED ISSUE #18: Multipart uploads buffer the full request body in memory

jbeckwith-oai:
Closes #18.

This replaces #19 with a current-main implementation.

## Root cause

The CLI opened file-backed values as readers, then synchronously copied those readers into a `bytes.Buffer` while constructing the multipart request option. Large uploads were therefore fully buffered before the request was sent, and CLI-owned file handles were not reliably closed.

## Design decision

Multipart forms containing uploads are intentionally single-attempt streams:

- encode lazily through `io.Pipe` with bounded memory
- set `WithMaxRetries(0)` because an arbitrary reader cannot be guaranteed to produce the same bytes twice
- return an explicit error for POST 307/308 responses that require replay
- do not create upload-sized temporary snapshots or duplicate descriptors
- retain `Content-Length` for trustworthy regular-file sizes; use chunked transfer for unknown-length readers

Scalar-only multipart forms remain buffered and retain the SDK's normal retry behavior.

This is a deliberate fail-closed boundary. Transparent upload replay would require either buffering/spooling the body or a first-class SDK replay contract. Reimplementing that retry state machine in CLI middleware introduced substantially more lifecycle and platform risk than the original memory/descriptor problem. A network failure during an upload now returns the original transport error after one attempt instead of retrying a body whose identity cannot be guaranteed.

## Correctness and cleanup

- source read failures abort the pipe without emitting a successful terminal multipart boundary
- exact-length regular-file readers report premature EOF if a file shrinks during the request
- owned upload files close exactly once across success, encoding failure, cancellation, close-before-read, and option-construction failure
- stdin and other unowned readers are not closed
- extensionless `@value` fallback occurs only for genuine not-found errors

## Impact

In the original local 128 MiB end-to-end reproduction, peak RSS dropped from approximately 348 MiB on `main` to approximately 26 MiB while the complete multipart payload reached the server.

Behavioral tradeoff: multipart requests containing uploads no longer receive automatic SDK retries. Scalar-only multipart requests are unchanged.

## Validation

- focused multipart regressions, including 307/308, 429/5xx single-attempt behavior, incomplete transport failures, Content-Length/chunked framing, premature EOF, cleanup, unavailable temp directories, low descriptor limits, and actual CLI receivers
- focused race suite
- stress runs for cancellation, incomplete transport failure, redirect rejection, and source shrinkage
- full `./scripts/test` native suite and Windows compile gate
- `./scripts/lint`
- `./scripts/build`
- `go vet ./...`


pkg/cmd/multipartbody.go
```diff
@@ -0,0 +1,142 @@
...
+func (b *multipartRequestBody) cleanup() error {
+	b.cleanupOnce.Do(func() {
+		b.cleanupErr = closeFileUploads(b.bodyMap)
+	})
+	return b.cleanupErr
+}
+
+func multipartRequestOptions(body *multipartRequestBody) []option.RequestOption {
+	return []option.RequestOption{
+		option.WithRequestBody(body.ContentType(), body),
```
HAYDEN-OAI, line 158: **[P1] Do not report unfollowed 307/308 redirects as successful uploads**

Passing this one-shot reader to `WithRequestBody` leaves `http.Request.GetBody == nil`. Go's HTTP client consequently refuses to follow POST 307/308 redirects, and the SDK treats the returned 3xx response as success. I reproduced `files create --file redirect-upload.txt --purpose assistants` against same-origin 307 and 308 responses: the parent sends two POSTs and returns the created file, while this commit sends only the initial POST, never reaches the upload endpoint, exits 0, and prints the redirect response as though the upload succeeded. This silently drops uploads through otherwise-supported custom gateways and canonicalizing proxies. Preserve replayability for seekable file-backed requests, or explicitly reject an unfollowed redirect rather than reporting success.

HAYDEN-OAI (commented):
I found four reproducible correctness and backward-compatibility regressions. Each inline finding was verified against both this exact PR commit and its parent with the actual CLI.

pkg/cmd/multipartbody.go
```diff
@@ -0,0 +1,142 @@
...
+	b.startOnce.Do(func() {
+		go b.encode()
+	})
+}
+
+func (b *multipartRequestBody) encode() {
+	defer close(b.done)
+
+	encodeErr := apiform.MarshalWithSettings(b.bodyMap, b.multipartWriter, b.encodingFormat)
+	writerCloseErr := b.multipartWriter.Close()
```
HAYDEN-OAI, line 117: **[P1] Do not emit a successful multipart terminator after a source read fails**

`multipartWriter.Close()` runs even when `MarshalWithSettings` already returned a file-read error, which emits a valid final MIME boundary for a truncated upload. Reproduced with `files create --file <directory> --purpose assistants`: the parent reports `is a directory` before making any HTTP request, whereas this commit sends an authenticated chunked POST containing a complete-looking zero-byte file part and terminal multipart boundary. A streaming multipart receiver parsed and persisted that zero-byte file even though the CLI subsequently exited with the read error. A mid-file disk or pipe failure can likewise commit corrupted remote data and make a retry duplicate the operation. On marshal/read failure, close the pipe with that error immediately and never write the successful final boundary; finalize the multipart writer only after encoding succeeds.

pkg/cmd/multipartbody.go
```diff
@@ -0,0 +1,142 @@
...
+	return b.cleanupErr
+}
+
+func multipartRequestOptions(body *multipartRequestBody) []option.RequestOption {
+	return []option.RequestOption{
+		option.WithRequestBody(body.ContentType(), body),
+		// A streamed multipart body cannot be replayed. Make the behavior
+		// explicit even though the SDK also avoids retrying bodies without
+		// an http.Request.GetBody function.
+		option.WithMaxRetries(0),
```
HAYDEN-OAI, line 159: **[P2] Preserve retries for multipart requests that contain no upload**

This unconditional override also disables retries when the form contains only small, replayable scalar fields. Reproduced the actual `videos create --prompt hello` command, which sends no file part, against a server returning `429` with `Retry-After: 0` followed by `200`: the parent issues two requests and successfully returns the queued video, while this commit issues one request and exits with `429 Too Many Requests`. Losing retries here is not the accepted tradeoff for streaming a non-replayable file; there is no file or reader to stream. Keep scalar-only multipart bodies buffered/replayable and preserve normal SDK retries, or otherwise provide replay support and disable retries only for genuinely non-replayable uploads.

pkg/cmd/flagoptions.go
```diff
@@ -492,22 +502,15 @@ func flagOptions(
...
 		}
 		encodingFormat := apiform.FormatBrackets
-		if err := apiform.MarshalWithSettings(bodyMap, writer, encodingFormat); err != nil {
-			return nil, err
-		}
-		if err := writer.Close(); err != nil {
-			return nil, err
-		}
-		options = append(options, option.WithRequestBody(writer.FormDataContentType(), buf))
+		body := newMultipartRequestBody(bodyMap, encodingFormat)
```
HAYDEN-OAI: **[P2] Retain Content-Length when every upload has a known size**

Routing every multipart form through an unknown-length pipe forces chunked transfer even for ordinary regular files with stat-able sizes. Against a supported `--base-url` upload gateway that requires `Content-Length`, the parent successfully uploads a regular file with `Content-Length: 414`, whereas this commit sends `Transfer-Encoding: chunked` with no length and fails with `411 Length Required`. This is a concrete compatibility regression independent of redirects and retries. Compute multipart framing plus regular-file sizes and set the request content length when all parts have known lengths, falling back to chunked transfer only for stdin, pipes, or other unknown-length sources.

pkg/cmd/multipartbody.go
jbeckwith-oai: Fixed in e48d568. Regular-file multipart uploads now install an HTTP GetBody factory that reopens the source with the same multipart boundary, while the middleware supplies the exact Content-Length. POST 307 and 308 therefore reach the redirected upload endpoint. Unknown/non-replayable readers now return an explicit redirect error instead of treating the 3xx as success. Covered by TestMultipartRequestOptionsFollow307And308ForRegularFiles and TestMultipartRequestOptionsReject307And308ForUnknownSources; focused race tests (20 iterations) and the full suite pass.

pkg/cmd/multipartbody.go
jbeckwith-oai: Fixed in e48d568. The encoder now calls CloseWithError immediately when MarshalWithSettings reports a source/read error and returns without closing the multipart writer, so it cannot emit a successful terminal MIME boundary. Directories are also rejected during file preparation before any request. Covered by TestMultipartRequestBodyPropagatesProducerErrors, TestMultipartRequestBodySourceErrorCannotBeParsedAsCompleteUpload, and TestOpenFileUploadRejectsDirectoriesBeforeRequest; the server-side regression verifies the truncated part cannot be parsed as a complete upload.

pkg/cmd/multipartbody.go
jbeckwith-oai: Fixed in e48d568. Multipart construction now uses a hybrid policy: forms with no reader/upload are encoded into a buffered byte body and do not override MaxRetries, preserving the SDK retry and replay behavior. Only genuinely non-replayable streaming sources disable retries. TestMultipartRequestOptionsRetryScalarOnlyForms reproduces videos create --prompt hello against 429 plus Retry-After: 0 followed by 200 and verifies two requests and success.

pkg/cmd/flagoptions.go
jbeckwith-oai: Fixed in e48d568. openFileUpload records regular-file size/reopen metadata. Multipart preparation computes exact framing with the real boundary and encoder, adds all known file sizes, and sets Request.ContentLength; stdin, pipes, and arbitrary readers remain unknown-length and chunked. TestMultipartRequestOptionsSetContentLengthForRegularFiles verifies the header equals the exact bytes sent, while TestMultipartRequestOptionsDisableRetriesForUnknownSources verifies unknown sources remain chunked. Full Linux, race, vet, build, lint, and Windows compile validation pass.

pkg/cmd/multipartbody.go
```diff
@@ -0,0 +1,378 @@
...
+		newBody := func() (io.ReadCloser, error) {
+			reopened, err := reopenMultipartBody(bodyMap)
+			if err != nil {
+				return nil, err
+			}
+			return newMultipartRequestBodyWithBoundary(reopened, encodingFormat, boundary)
+		}
+		options = append(options, option.WithMiddleware(func(req *http.Request, next option.MiddlewareNext) (*http.Response, error) {
+			req.ContentLength = contentLength
+			req.GetBody = newBody
```
HAYDEN-OAI: **[P2] Preserve SDK retries for replayable regular-file uploads**

This middleware assigns `GetBody` only to the attempt-local request. In openai-go v3.51.0, middleware receives `cfg.Request.Clone(...)`, but `shouldRetry` and retry-body reconstruction inspect `cfg.Request.GetBody`, which remains nil. Consequently every ordinary file-backed upload silently loses the existing 408/409/429/5xx and connection-error retries despite being replayable. Reproduced the actual `files create --file upload.txt --purpose assistants` CLI against `429` plus `Retry-After: 0` followed by `200`: the base sends two POSTs and succeeds, while this head sends one complete Content-Length-bearing POST and exits with `429 Too Many Requests`. This is separate from the already-fixed scalar-only retry issue. Install the replay factory on the SDK-owned request before its retry decision, or otherwise arrange retries at that layer.

pkg/cmd/flagoptions.go
```diff
@@ -562,6 +571,32 @@ func (f fileUpload) Close() error {
...
+	}
+	file, err := os.Open(f.sourcePath)
+	if err != nil {
+		return fileUpload{}, err
+	}
+	info, err := file.Stat()
+	if err != nil {
+		return fileUpload{}, errors.Join(err, file.Close())
+	}
+	if !info.Mode().IsRegular() || info.Size() != f.size {
```
HAYDEN-OAI: **[P2] Ensure redirect replay uploads the original file contents**

Reopening the pathname and checking only that it is regular and the same size accepts a different file after an atomic replacement or symlink retarget. Reproduced the actual CLI with a gateway issuing `307` after the first upload while the source path changes from `benign-data` to equal-length `secret-data`: the base sends `benign-data` on both POSTs, but this head sends `benign-data` initially and silently uploads `secret-data` after the redirect; both commands report success. Rotated/generated artifacts can therefore be corrupted, and pathname substitution can disclose unintended local data. Preserve a stable descriptor/snapshot for replay, or reject changes in file identity/content instead of validating size alone.

pkg/cmd/flagoptions.go
```diff
@@ -578,6 +620,9 @@ func openFileUpload(path string) (fileUpload, error) {
 		Reader:      file,
 		filename:    filepath.Base(path),
 		contentType: contentType,
+		sourcePath:  path,
+		size:        info.Size(),
+		knownSize:   info.Mode().IsRegular(),
```
HAYDEN-OAI: **[P2] Do not infer a trustworthy file length from `IsRegular` alone**

Linux procfs/sysfs pseudo-files report regular mode even when their stat size differs from their readable contents. Reproduced `files create --file /proc/version --purpose assistants`: `/proc/version` reports size 0 but yields 187 bytes; the base succeeds with `Content-Length: 552`, while this head advertises 365, sends an incomplete multipart form, and fails with `http: ContentLength=365 with Body length 552`. `/proc/cpuinfo`, `/proc/meminfo`, and sysfs files have the same problem, including overreported sizes. Avoid marking virtual files known-size/replayable solely from their mode; use chunked streaming unless the reported length is trustworthy.

HAYDEN-OAI (commented):
Verified that the earlier redirect handling, source-read failures, scalar-only retries, and Content-Length findings were addressed. I found three additional, independently reproduced regressions in regular-file retry/replay and virtual-file length handling; details are inline.

pkg/cmd/multipartbody.go
jbeckwith-oai: Fixed in 521f01c. Replayable regular-file forms now give openai-go a concrete buffered replay marker so it installs GetBody on the SDK-owned canonical request; attempt middleware replaces that marker with the real streamed multipart body, preserving the SDK’s native retry loop, backoff, and retry-count headers while retaining exact Content-Length. TestMultipartRequestOptionsRetryRegularFiles reproduces files create against 429 + Retry-After: 0 then 200 and verifies two complete identical uploads, stable Content-Length, and SDK retry counts 0/1. TestMultipartRequestOptionsCloseRegularFilesAfterFinalRetry verifies all three default attempts and terminal descriptor cleanup. Full tests and normal Windows compile pass; full native race, vet, lint, and build pass.

pkg/cmd/flagoptions.go
jbeckwith-oai: Fixed in 521f01c. Regular uploads are opened once and replayed with independent SectionReaders over that stable *os.File descriptor; redirects and retries no longer reopen or trust the pathname, so atomic replacement or symlink retargeting cannot substitute another file. TestMultipartRequestOptionsRedirectReplayUsesOpenedFile replaces benign-data with equal-length secret-data during a 307 and verifies both POSTs still contain benign-data while the pathname contains secret-data afterward. Full tests and normal Windows compile pass; full native race, vet, lint, and build pass.

pkg/cmd/flagoptions.go
jbeckwith-oai: Fixed in 521f01c. A regular-looking file is considered known-length only after ReadAt probes validate the reported last byte and EOF without changing its offset. Sources whose stat size is not trustworthy—including procfs/sysfs-style files—remain unknown-length, non-replayable streams and use chunked transfer. TestOpenFileUploadStreamsProcfsWithUnknownLength reproduces /proc/version and verifies ContentLength -1, chunked transfer, nonempty contents, and cleanup; TestOpenFileUploadTreatsEmptyRegularFileAsKnownLength covers the real empty-file boundary. Full tests and normal Windows compile pass; full native race, vet, lint, and build pass.

pkg/cmd/multipartbody.go
```diff
@@ -0,0 +1,527 @@
...
+
+func (b *multipartRequestBody) encode() {
+	defer close(b.done)
+
+	if err := apiform.MarshalWithSettings(b.bodyMap, b.multipartWriter, b.encodingFormat); err != nil {
+		// A final multipart boundary asserts that every part completed. Abort the
+		// pipe immediately on a source error instead of making a truncated file
+		// look like a valid zero-byte upload to the receiver.
+		_ = b.writer.CloseWithError(err)
+		_ = b.cleanup()
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 8724259. The encoder-error path now closes the pipe with errors.Join(encodeErr, cleanupErr), preserving the original source failure while exposing every owned-upload close failure through the body reader. TestMultipartRequestBodyJoinsCleanupErrorAfterSourceError failed first on 521f01c and now verifies both errors are in the returned chain and cleanup runs once. The complete multipart race suite passed 50 iterations; full native tests/race, normal Windows compile, vet, lint, build, diff-check, and thermo-nuclear review pass.

pkg/cmd/multipartbody.go
```diff
@@ -0,0 +1,527 @@
...
+	body, err := attempt.newBody()
+	if err != nil {
+		return nil, errors.Join(err, r.close())
+	}
+	req.Body = body
+	req.ContentLength = r.contentLength
+	req.GetBody = attempt.newBody
+
+	res, requestErr := next(req)
+	attemptErr := attempt.close()
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 8724259. Replay bodies now read through duplicated per-body file handles. A successful terminal response seals the attempt and releases transport-owned bodies to finish naturally; it closes only the stable request anchor, which cannot truncate those duplicates. Error, retry, and cancellation paths still abort their bodies. TestMultipartRequestOptionsLetEarlySuccessfulUploadFinish failed first on 521f01c and now uses an early-response HTTP doer to verify the API can return before body consumption resumes, after which the exact Content-Length is fully read without error and the anchor is closed. Full native tests/race, normal Windows compile, vet, lint, build, diff-check, and thermo-nuclear review pass.

pkg/cmd/multipartbody.go
```diff
@@ -0,0 +1,527 @@
...
+		return nil
+	}
+	a.closed = true
+	bodies := append([]*multipartRequestBody(nil), a.bodies...)
+	a.mu.Unlock()
+
+	var errs []error
+	for _, body := range bodies {
+		errs = append(errs, body.Close())
+		<-body.done
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 8724259. Each replay/redirect body now owns a duplicated OS file handle (Unix Dup; Windows DuplicateHandle), so abort can close that attempt’s handles independently while retaining the stable anchor for a future retry. Attempt abort seals and closes bodies but never waits on encoder completion, so an uninterruptible ReadAt cannot block cancellation/error return. TestMultipartReplayAttemptAbortDoesNotWaitForBlockedReader failed first on 521f01c and now proves abort returns before a manually blocked source is released. Full native tests/race, normal Windows compile, vet, lint, build, diff-check, and thermo-nuclear review pass.

pkg/cmd/multipartbody.go
```diff
@@ -0,0 +1,527 @@
...
+	req.Body = body
+	req.ContentLength = r.contentLength
+	req.GetBody = attempt.newBody
+
+	res, requestErr := next(req)
+	attemptErr := attempt.close()
+	if req.Context().Err() != nil ||
+		!shouldRetryReplayableMultipart(res) ||
+		attemptIndex >= replayableMultipartMaxRetries {
+		return res, errors.Join(requestErr, attemptErr, r.close())
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 8724259. Every SDK attempt installs a context finalizer; when a retry remains pending, that finalizer stays armed through Retry-After/backoff and closes the stable request anchors if cancellation prevents the next middleware attempt. Starting the next attempt disarms the previous watcher, and terminal cleanup remains idempotent. TestMultipartRequestOptionsCloseRegularFilesWhenBackoffCanceled failed first on 521f01c, cancels only after the attempt middleware has returned, verifies one POST/context.Canceled, and verifies the anchor closes. Full native tests/race, normal Windows compile, vet, lint, build, diff-check, and thermo-nuclear review pass.

pkg/cmd/flagoptions.go
```diff
@@ -551,17 +558,107 @@ type fileUpload struct {
...
+	}
+	if !info.Mode().IsRegular() || info.Size() != s.size {
+		return nil, fmt.Errorf("upload %q changed while being replayed", s.file.Name())
+	}
+	file, err := duplicateFile(s.file)
+	if err != nil {
+		return nil, fmt.Errorf("duplicate upload %q: %w", s.file.Name(), err)
+	}
+	return &sectionReadCloser{
+		SectionReader: io.NewSectionReader(file, 0, s.size),
```
HAYDEN-OAI: **[P1] Fail closed when a replayable file shrinks while streaming**

`io.NewSectionReader(file, 0, s.size)` returns ordinary `EOF` when the underlying file is truncated before `s.size` bytes are read. `apiform`'s `io.Copy` treats that as success, so the encoder emits a valid final multipart boundary around the shortened file. I reproduced the actual `files create` command against a streaming receiver that truncates a 32 MiB source after upload starts: the pre-PR base committed all 33,554,432 bytes, while this head advertised `Content-Length: 33554809` but sent a complete-looking multipart form whose committed file part contained only 1,081,344 bytes. The CLI eventually fails with `stat ...: file already closed`, but the receiver has already persisted corrupted data. Wrap replay readers in an exact-length reader that converts premature EOF to `io.ErrUnexpectedEOF`, ensuring encoding aborts without emitting the final boundary.

pkg/cmd/flagoptions.go
```diff
@@ -551,17 +558,107 @@ type fileUpload struct {
...
+	closeOnce sync.Once
+	closeErr  error
+}
+
+func (s *replayableFileSource) ReplayReader() (io.ReadCloser, error) {
+	info, err := s.file.Stat()
+	if err != nil {
+		return nil, err
+	}
+	if !info.Mode().IsRegular() || info.Size() != s.size {
```
HAYDEN-OAI: **[P2] Reject same-inode content changes before replay**

Checking only regular-file mode and size does not detect an equal-length in-place rewrite, and duplicated descriptors still observe changes to their shared underlying inode. I reproduced the actual `files create` CLI with a server that changes `benign-data` to `secret-data` in place after the first POST: for each of 307, 308, and 429, the pre-PR base sends `[benign-data, benign-data]`, while this head sends `[benign-data, secret-data]` with the same inode, unchanged `Content-Length`, and a successful exit. This bypasses the fixed pathname-replacement case and can corrupt replayed uploads or disclose newly written local content. Reject source mutations using trustworthy change metadata/content identity before each replay, or preserve an immutable snapshot.

HAYDEN-OAI (commented):
Re-reviewed exact head `872425944c457cd1bbc71d4869e84e6b5aaff431` after the prior redirect, retry, virtual-file framing, and cleanup feedback was addressed. Direct comparisons against the pre-PR base confirm two remaining upload integrity issues; details and reproductions are inline.

pkg/cmd/flagoptions.go
jbeckwith-oai: Fixed in af8013d49bdc4dd06563160b9eb33edc60513e77. Replayable regular files are now copied once into an exact-size anonymous on-disk snapshot, and every body reads through exactLengthReader. A premature EOF becomes io.ErrUnexpectedEOF, so the existing encoder error path closes with error and never calls multipart.Writer.Close or emits the terminal MIME boundary. TestMultipartReplayPrematureEOFDoesNotCommitFinalBoundary drains the receiver and verifies the form is not committed; TestFilesCreateCLISnapshotsBeforeSourceShrinks exercises the actual files create CLI while truncating the original after the first uploaded byte and verifies the complete snapshotted payload arrives. Validation: focused race/stress x20, full native Go tests and race phase, go vet, lint, build, Windows cross-compile, and the sealed security review all passed.

pkg/cmd/flagoptions.go
jbeckwith-oai: Fixed in af8013d49bdc4dd06563160b9eb33edc60513e77. Redirects and retries no longer read the caller inode: openFileUpload creates one bounded disk snapshot (unlinked immediately on Unix; exclusive delete-on-close on Windows), and each replay duplicates that immutable snapshot handle while retaining Content-Length. TestFilesCreateCLIUsesImmutableReplayContent runs the actual files create CLI, rewrites the original in place with equal-length content after request one, and verifies both attempts retain the original bytes for 307, 308, and 429. Snapshot construction also enforces the original exact length and closes every source and snapshot on failure. Validation: focused race/stress x20, full native Go tests and race phase, go vet, lint, build, Windows cross-compile, and the sealed security review all passed.

pkg/cmd/multipartfile_windows.go
```diff
@@ -0,0 +1,53 @@
...
+	path := placeholder.Name()
+	if err := placeholder.Close(); err != nil {
+		return nil, errors.Join(err, os.Remove(path))
+	}
+	if err := os.Remove(path); err != nil {
+		return nil, err
+	}
+	file, err := os.OpenFile(
+		path,
+		os.O_RDWR|os.O_CREATE|os.O_EXCL|windows.O_FILE_FLAG_DELETE_ON_CLOSE,
```
HAYDEN-OAI: **[P1] Actually delete Windows upload snapshots on the supported Go 1.25 toolchain**

The release workflow installs the `go 1.25.0` version declared in `go.mod`, but that version's Windows `os.OpenFile` → `syscall.Open` silently drops the high `O_FILE_FLAG_DELETE_ON_CLOSE` bit. I cross-compiled this exact head with Go 1.25 and 1.26 and executed their unmodified `syscall.Open` implementations with these exact flags: `CreateFile` receives `0x00200080` on Go 1.25 (no delete flag), versus `0x04200080` on Go 1.26. Because successful cleanup only calls `file.Close()`, every Windows regular-file upload leaves a complete, named plaintext `openai-multipart-upload-*` copy in `%TEMP%` indefinitely, exposing sensitive uploaded data and accumulating disk. Call `windows.CreateFile` with an explicitly applied delete-on-close flag, reliably remove the pathname after the final duplicate closes, or require a Go version that actually forwards this flag.

pkg/cmd/multipartfile.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+	if !hasTrustworthyFileSize(file, info) {
+		return fileUpload{
+			Reader:      file,
+			filename:    filepath.Base(path),
+			contentType: contentType,
+		}, nil
+	}
+
+	source, err := snapshotReplayableFile(file, info.Size(), path)
```
HAYDEN-OAI: **[P2] Do not require upload-sized writable temporary storage before the first request**

This unconditional snapshot makes every regular-file upload depend on a writable temporary filesystem with at least the full logical file size, even when no redirect or retry occurs. With `TMPDIR` backed by a real 64 KiB tmpfs, the pre-PR CLI successfully sent a complete authenticated 256 KiB file upload, while this head exited `no space left on device` without making any request; a sparse 256 KiB file occupying only 4 KiB on its original filesystem failed identically because the snapshot materializes its holes. A missing/read-only `TMPDIR` also breaks uploads that previously succeeded, including zero-byte files. Preserve a streaming first-attempt fallback when a replay snapshot cannot be created, explicitly disabling replay if necessary, instead of making full-size scratch storage mandatory.

pkg/cmd/multipartfile.go
```diff
@@ -0,0 +1,232 @@
...
+func openFileUpload(path string) (fileUpload, error) {
+	file, err := os.Open(path)
+	if err != nil {
+		return fileUpload{}, err
+	}
+	info, err := file.Stat()
+	if err != nil {
+		return fileUpload{}, errors.Join(err, file.Close())
+	}
+	if info.IsDir() {
```
HAYDEN-OAI, line 117: **[P2] Propagate preparation failures for existing extensionless `@name` uploads**

The new early directory and snapshot errors flow into `embedFilesValue`'s existing extensionless-name fallback, which currently treats *every* `openFileUpload` error as evidence that `@name` is a literal username. With an existing directory named `artifactdir`, the actual `videos create --prompt hello --input-reference @artifactdir` command previously failed with `read artifactdir: is a directory` and sent no request; this head instead sends an authenticated request containing the scalar `input_reference=@artifactdir` and reports success. With an existing readable file named `artifact` and an unwritable temporary directory, the base uploads its 27-byte file part but this head likewise silently sends the literal `@artifact` instead. Retain the literal fallback only for a genuinely nonexistent target (`errors.Is(err, fs.ErrNotExist)`); propagate directory, permissions, snapshot, and source-read failures.

pkg/cmd/multipartbody.go
```diff
@@ -0,0 +1,584 @@
...
+	}
+	if int64(framing.Len()) > math.MaxInt64-uploadBytes {
+		return 0, errors.New("multipart content length overflows int64")
+	}
+	return int64(framing.Len()) + uploadBytes, nil
+}
+
+func replayMultipartBody(bodyMap map[string]any) (map[string]any, error) {
+	value, err := transformFileUploads(bodyMap, func(upload fileUpload) (fileUpload, error) {
+		return upload.replay()
```
HAYDEN-OAI: **[P2] Avoid eagerly doubling file descriptors for multi-file uploads**

`replayMultipartBody` duplicates every upload handle before streaming begins while retaining all stable snapshot anchors, so an N-file form needs roughly 2N file descriptors before it can even open its network socket. The `images edit` command explicitly supports 16 images: against the same authenticated HTTP receiver, the pre-PR CLI successfully uploaded all 16 images with `RLIMIT_NOFILE=23`, while this head failed without sending any request for every limit from 23 through 38 and required 39 descriptors. Depending on the limit, users see either `socket: too many open files` or the misleading `stat ...: file already closed`. Open and release replay handles lazily per multipart part, share a bounded backing handle, or otherwise avoid allocating one additional live descriptor for every file before dialing.

HAYDEN-OAI (commented):
Re-reviewed exact head `af8013d49bdc4dd06563160b9eb33edc60513e77` against the pre-PR base after excluding all previously resolved findings. I found four newly introduced, independently validated runtime/security/resource regressions; the Linux cases were reproduced with the actual CLI and authenticated local HTTP receivers, and the Windows case was verified against exact-head release-toolchain binaries and the unmodified Go syscall implementation.

pkg/cmd/multipartfile_windows.go
jbeckwith-oai: Fixed at 7e4a444fad016106c162dd4e9f794cb4dd16caa1. Windows replay capture now calls `windows.CreateFile` directly, so `FILE_FLAG_DELETE_ON_CLOSE` reaches Win32 on the supported Go 1.25 toolchain instead of passing through the flag-dropping `os.OpenFile` path. Creation uses `CREATE_NEW`, only `FILE_SHARE_DELETE`, and a protected current-user-only DACL; the final duplicated-handle close removes the pathname. `TestReplaySnapshotIsDeletedAfterFinalHandleCloses` covers final-handle deletion, and `TestReplaySnapshotHasProtectedCurrentUserDACL` verifies the protected single ACE, no inherited ACEs, and that a second pathname open is denied. The package cross-compiles with Go 1.25 for windows/amd64; full tests/race, lint, vet, build, GoReleaser packaging, thermo review, and a fresh complete exact-range security scan (0 findings) pass.

pkg/cmd/multipartfile.go
jbeckwith-oai: Fixed at 7e4a444fad016106c162dd4e9f794cb4dd16caa1 by replacing mandatory per-file preflight snapshots with a streaming-first hybrid. The initial request reads the original source directly and opportunistically captures the exact encoded multipart body into one lazily created replay file. Snapshot create/write failure does not fail or delay that first upload; it closes the partial capture and disables 307/308/SDK retry replay so no non-identical attempt is sent. `TestFilesCreateCLIStreamsWhenSnapshotDirectoryUnavailable` covers nonempty and zero-byte files, `TestFilesCreateCLIDisablesReplayWhenSnapshotDirectoryUnavailable` covers 307 and 429, and `TestFilesCreateCLIStreamsSparseFileWhenSnapshotQuotaExceeded` reproduces a 256 KiB sparse upload under a 64 KiB snapshot quota. Full tests/race, lint, vet, build, GoReleaser packaging, thermo review, and the fresh exact-range security scan pass.

pkg/cmd/multipartfile.go
jbeckwith-oai: Fixed at 7e4a444fad016106c162dd4e9f794cb4dd16caa1. The extensionless `@name` literal fallback now runs only when `errors.Is(err, fs.ErrNotExist)` and the value did not explicitly look like a file; directory, permission, stat, source, and replay-preparation errors propagate. Replay snapshot availability is no longer part of file preparation, so an existing extensionless file still streams even when temp capture cannot be created. `TestVideosCreateCLIPropagatesExtensionlessDirectoryError` verifies the actual CLI fails before any request, and `TestVideosCreateCLIUploadsExtensionlessFileWhenSnapshotDirectoryUnavailable` verifies the existing file is uploaded rather than sent as the scalar literal. Full tests/race, lint, vet, build, packaging, thermo review, and the fresh exact-range security scan pass.

pkg/cmd/multipartbody.go
jbeckwith-oai: Fixed at 7e4a444fad016106c162dd4e9f794cb4dd16caa1. Request construction no longer duplicates every upload descriptor. Each selected file is opened once and streamed lazily on the first attempt; one lazily created encoded-body snapshot (not N per-file snapshots/duplicates) backs any redirect or SDK retry, and replay bodies duplicate only that single bounded handle. `TestImagesEditCLIDialsBeforeOpeningReplayDescriptors` runs the actual 16-image CLI under `RLIMIT_NOFILE=23` and verifies the receiver gets all files, while the capture/retry lifecycle tests cover closure and cancellation. Full tests and full race, current and Go 1.25 Windows cross-compiles, lint, vet, build, GoReleaser packaging, thermo review, and the fresh complete exact-range security scan pass.

pkg/cmd/multipartbody.go
```diff
@@ -0,0 +1,624 @@
...
+	req.GetBody = attempt.newBody
+
+	res, requestErr := next(req)
+	requestCanceled := req.Context().Err() != nil
+	replayAvailable := r.source.Ready()
+	if res != nil && shouldRetryReplayableMultipart(res) && !replayAvailable {
+		// The first upload is still valid when the optional replay snapshot
+		// cannot be created. Tell openai-go not to issue another network
+		// attempt that this request cannot reproduce faithfully.
+		res.Header.Set("x-should-retry", "false")
```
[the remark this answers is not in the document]
jbeckwith-oai: You're right about the behavior at `7e4a444`: once the first transport attempt returned `res == nil` before capture completed, the CLI middleware had no supported way to veto the SDK retry without replacing the useful transport error.

Rather than add another retry-state workaround, we stepped back and removed CLI-managed replay entirely in `27f98fe`. Multipart requests containing an upload now use the smaller option explicitly identified in issue #18:

- stream the body once through `io.Pipe`
- set `WithMaxRetries(0)` for every upload-bearing multipart request
- return an explicit error for 307/308 responses that require replay
- keep scalar-only multipart forms buffered and normally retryable
- keep `Content-Length` for trustworthy regular files and chunk unknown-length sources

This makes the nil-response case deterministic: the SDK stops after attempt zero and preserves the original transport error. `TestMultipartRequestOptionsPreserveIncompleteTransportError` reproduces a transport failure before the body is read, asserts one `Do` call, asserts `errors.Is` against the original network error, and verifies source cleanup.

The single-attempt policy is deliberate. CLI upload sources can be mutable files, stdin, pipes, or arbitrary readers, so the CLI cannot promise byte-identical replay. Providing that guarantee would require full buffering/spooling or a first-class SDK replay contract; emulating it in middleware was outside the original bounded-memory/descriptor-leak fix and created the lifecycle/platform complexity seen in the previous revisions. Failing closed after one streamed attempt is therefore the correct ownership boundary for this PR.

Validation on the pushed head: focused race tests; 50x stress coverage for cancellation, incomplete transport errors, source shrinkage, and 307/308 rejection; full native tests plus Windows compile via `./scripts/test`; lint; build; and `go vet ./...`.

pkg/cmd/multipartbody_test.go
```diff
@@ -0,0 +1,620 @@
...
+	bytesRead atomic.Int64
+}
+
+func (r *countingReader) Read(p []byte) (int, error) {
+	if r.remaining == 0 {
+		return 0, io.EOF
+	}
+	if int64(len(p)) > r.remaining {
+		p = p[:r.remaining]
+	}
```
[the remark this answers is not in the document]
jbeckwith-oai: No code change is needed in this test helper. As specified for [index expressions](https://go.dev/ref/spec#Index_expressions), a slice index may have any integer type; it is not restricted to `int`. This `int64` index mirrors the valid production expression, and the preceding bounds check keeps it within `len(p)`.

Exact head `27f98fe3b2092916e4117bc7a4c9f47baf6b3d5e` compiled successfully in the hosted `test`, `build artifacts`, and `build` jobs, plus the local native and Windows compile gates. Resolving as a false positive.

pkg/cmd/multipartfile.go
```diff
@@ -0,0 +1,145 @@
...
+	remaining int64
+}
+
+func (r *exactLengthReader) Read(p []byte) (int, error) {
+	if r.remaining == 0 {
+		return 0, io.EOF
+	}
+	if int64(len(p)) > r.remaining {
+		p = p[:r.remaining]
+	}
```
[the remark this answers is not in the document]
jbeckwith-oai: No code change is needed here. The Go specification for [index expressions](https://go.dev/ref/spec#Index_expressions) requires the index to be of integer type or untyped; it does not require the concrete type to be `int`. `r.remaining` is `int64`, so `p[:r.remaining]` is valid, and the preceding comparison ensures it is no greater than `len(p)`.

This unchanged exact head, `27f98fe3b2092916e4117bc7a4c9f47baf6b3d5e`, compiled successfully in the hosted `test`, `build artifacts`, and `build` jobs, as well as the local native and Windows compile gates. Resolving as a false positive.

pkg/cmd/multipartbody.go
```diff
@@ -0,0 +1,344 @@
...
+		option.WithMiddleware(streamedMultipartMiddleware(contentLength)),
+	}, nil
+}
+
+func streamedMultipartMiddleware(contentLength int64) option.Middleware {
+	return func(req *http.Request, next option.MiddlewareNext) (*http.Response, error) {
+		if contentLength >= 0 {
+			req.ContentLength = contentLength
+		}
+		res, err := next(req)
```
HAYDEN-OAI, line 181: **[P2] Close stalled streamed uploads when their request context is canceled**

Canceling an in-flight upload can hang indefinitely when an owned FIFO or other unknown-length source stops producing data. Here next(req) waits for the HTTP transport write loop, which waits for multipartRequestBody.Read; the encoder is simultaneously blocked in os.File.Read on the still-open FIFO. The SDK only closes req.Body after next(req) returns, so cancellation never reaches multipartRequestBody.Close() to release the source, creating a circular wait. I reproduced the exact base and head with the actual files create command, an authenticated HTTP receiver, and a FIFO producer held open for four seconds: after the request began, the base returned context canceled in 103 µs, while this head began streaming after 3 ms but remained blocked for 3.998 seconds after cancellation, with both FIFO descriptors live, until the producer finally closed; a permanently stalled producer hangs forever. Close the streamed body/owned upload sources from an independent request-context cancellation watcher while next(req) is still executing, and stop the watcher when the attempt finishes.

HAYDEN-OAI (commented):
Independently re-reviewed exact head 27f98fe3b2092916e4117bc7a4c9f47baf6b3d5e against pre-PR base d14387a46865b8d227d3a1d607dfd40143904047, including the resolved findings and direct base-versus-head CLI reproductions. One newly introduced cancellation/resource-lifecycle regression in the single-attempt streaming path is documented inline.

pkg/cmd/multipartbody.go
jbeckwith-oai: Fixed at `1a74fd7630d7deb37dd3b17f4577cbf5dae8d870`. The streamed-multipart middleware now registers a request-context cancellation callback before `next(req)` and closes `req.Body` independently while the transport is running; because that body owns the multipart source, a stalled FIFO read is released immediately. The callback is stopped after the attempt and, if it already started, the middleware waits for it to finish so no cancellation goroutine can outlive the attempt.

The fail-first regression uses the real `files create` CLI path, an authenticated `httptest` receiver, and a FIFO producer that deliberately remains open: before the fix the canceled request stayed blocked until the producer was released; after the fix it returns `context.Canceled` immediately, the still-open producer gets `EPIPE`, and the receiver observes an incomplete multipart stream. Additional tests prove the source closes exactly once and canceling after a completed attempt does not trigger a late close. Validation passed: focused tests, 50x `-race` stress, full `./scripts/test` including Windows compile, `./scripts/lint`, `./scripts/build`, and `go vet ./...`.

HAYDEN-OAI (approved):
Reviewed the exact requested head and found no substantive correctness, security, compatibility, or reliability issues. Independently verified stalled FIFO/stdin cancellation, truncated-source integrity, scalar retries, explicit non-replayable redirects, accurate framing, bounded memory, descriptor limits, and the absence of upload snapshots.

PR #90, opened 2026-08-18
fix(multipart): handle typed nil readers consistently

### LINKED ISSUE #89: Multipart encoder can panic on typed nil io.Reader pointers
## Summary

`internal/apiform` detects `io.Reader` implementations before it applies the encoder's existing nil-pointer handling. A typed nil pointer whose type implements `io.Reader` can therefore be routed into `io.Copy`, which invokes `Read` on a nil receiver and can panic.

The encoder already defines nil pointers as empty form fields, so reader detection currently bypasses an intended invariant.

## Reproduction

On current `main` (`d082a010f7c6cacf407d8a1581446a7857f9f1bb`), a minimal reader type demonstrates the problem:

```go
type panicReader struct{}

func (*panicReader) Read([]byte) (int, error) {
    panic("Read called on nil receiver")
}

var reader *panicReader

var buf bytes.Buffer
writer := multipart.NewWriter(&buf)
_ = Marshal(map[string]any{"file": reader}, writer)
```

Because `*panicReader` implements `io.Reader`, `encodeValue` checks:

```go
if t.Implements(reflect.TypeOf((*io.Reader)(nil)).Elem()) {
    return e.encodeReader(key, val, writer)
}
```

before reaching the pointer branch:

```go
case reflect.Pointer:
    if val.IsNil() || !val.IsValid() {
        return writer.WriteField(key, "")
    }
```

The nil pointer is therefore converted to an `io.Reader` interface and passed to `io.Copy`, which calls its `Read` method.

## Expected behavior

A typed nil pointer should follow the same nil-pointer semantics regardless of whether its type happens to implement `io.Reader`: encode an empty field and do not invoke methods on the nil receiver.

## Suggested fix

Check nil pointer/interface values before reader-interface detection. Keep reader handling unchanged for non-nil values.

Add a regression using a typed nil reader whose `Read` method panics if invoked, and verify multipart encoding completes without a panic and emits an empty field.

## Impact

This is robustness and request-serialization correctness. A legitimate typed nil value can currently turn request construction into a process panic instead of being handled by the encoder's existing null-value path.

sylvesterkaczmarek:
## Summary

Normalize typed-nil reader semantics across multipart encoding and request transport selection so typed-nil readers are treated consistently as empty scalar fields.

Fixes #89.

## Root cause

`internal/apiform.encodeValue` detected `io.Reader` capability before applying nil-pointer semantics. A typed-nil pointer therefore retained a concrete reader type and could reach `io.Copy`.

The request layer had the same semantic mismatch in a different place: `inspectMultipartBody` classified the same typed-nil reader as an upload even after the encoder treated it as an empty field. That unnecessarily selected one-shot streaming, disabled retries, and made an otherwise replayable scalar form reject 307/308 redirects.

## Fix

- unwrap interface layers before reader detection in the multipart encoder and treat a resulting nil pointer as an empty field;
- classify typed-nil readers as scalar values before choosing buffered versus streamed multipart transport;
- keep typed-nil readers out of the unknown-size-reader path used for known-length upload framing, so they can coexist with regular file uploads without breaking `Content-Length` calculation;
- leave non-nil readers and real uploads on the existing streamed, non-retryable path.

## Regression coverage

The tests cover:

- a concrete typed-nil reader stored in `map[string]any`;
- a typed-nil reader stored in `map[string]io.Reader`;
- a genuinely nil `io.Reader` interface;
- the established empty-field multipart representation;
- transport classification of a typed-nil reader as scalar/replayable;
- retry after an HTTP 429 while preserving the empty scalar field and buffered `Content-Length`;
- replay across both HTTP 307 and 308 redirects;
- a known-length file upload combined with a typed-nil optional reader, including exact-length streamed transport.

## Validation

The earlier encoder-focused revision passed focused tests, race tests, repository compile-only tests, `go mod verify`, and lint locally. The transport follow-up is covered by the new request-level regressions above.

Fresh upstream CI, CodeQL, and Castiron runs are currently gated with `action_required` pending external-contributor approval.

## Risk

Low. The behavior change is limited to typed-nil reader values. Non-nil readers and real uploads retain the existing streamed path with retries disabled; scalar-only forms remain buffered and replayable.


jbeckwith-oai:
I followed up on the typed-nil review and pushed e6eeec6 via maintainer edits. The original shallow nil check still allowed a typed-nil pointer wrapped in an `io.Reader`-typed map or slice to reach `io.Copy`, because the outer interface itself was non-nil and implemented `io.Reader`. The encoder now unwraps interface layers before reader detection, checks the concrete pointer for nil, and removes the redundant interface/pointer branches. Regression coverage now includes `map[string]any`, `map[string]io.Reader` containing a typed-nil pointer, and a genuinely nil reader interface. Local validation passed: focused tests, package race tests, repository compile-only tests, `go mod verify`, diff checks, and the lint build (with VCS stamping disabled for this linked-worktree environment).

HAYDEN-OAI (approved):
Reviewed typed-nil reader handling across interface/pointer unwrapping, multipart field encoding, reader dispatch, header validation, and streamed request ownership/error propagation. No substantive correctness, security, or compatibility issues found.

internal/apiform/encoder.go
```diff
@@ -49,6 +49,19 @@ func (e *encoder) encodeValue(key string, val reflect.Value, writer *multipart.W
...
+	// Unwrap interfaces before detecting io.Reader. A non-nil interface can
+	// contain a typed nil pointer, which must retain the encoder's empty-field
+	// semantics instead of being passed to io.Copy.
+	for val.Kind() == reflect.Interface {
+		if val.IsNil() {
+			return writer.WriteField(key, "")
+		}
+		val = val.Elem()
+	}
+	if val.Kind() == reflect.Pointer && val.IsNil() {
```
jbeckwith-oai, line 61: **[P2] Apply the same typed-nil semantics before choosing multipart transport.** This branch correctly serializes a typed-nil reader as an empty scalar field, but `pkg/cmd/multipartbody.go:inspectMultipartBody` still sees the non-nil `io.Reader` interface and marks it as `hasUpload`. As a result, `multipartRequestOptions` unnecessarily selects one-shot streaming, applies `option.WithMaxRetries(0)`, and rejects otherwise replayable 307/308 redirects. I verified the mismatch end to end with a synthetic HTTP 429 response: the empty-field request made one attempt and failed instead of retrying. Please update the upload classifier to recognize typed-nil readers as scalars and add a request-level retry regression.

jbeckwith-oai (changes requested):
Requesting changes because typed-nil reader normalization is inconsistent between multipart serialization and request transport selection. The encoder now writes a typed-nil reader as an empty scalar field, but `inspectMultipartBody` still classifies it as an upload and `multipartRequestOptions` therefore disables retries and selects a one-shot streamed body.

I reproduced this with a synthetic `httptest` server: a multipart request containing `map[string]any{"file": (*reader)(nil)}` received HTTP 429 and made only one attempt instead of retrying as a scalar-only form. Please normalize typed-nil readers consistently in the upload classifier and add an end-to-end request regression covering retry behavior. Redirect/replayability and Content-Length behavior should remain consistent with an empty scalar field as well.

The current-head CI, CodeQL, and Castiron custom-code workflows are also all stopped with `action_required`, so those checks still need to run successfully before merge.

internal/apiform/encoder.go
sylvesterkaczmarek: Fixed. The transport classifier now applies the same typed-nil reader semantics as the encoder, so typed-nil readers stay on the buffered scalar path instead of being treated as uploads. I also updated the known-length framing path so a typed-nil optional reader can coexist with a real file upload.

Added request-level regressions covering the reported 429 retry case, replay across both 307 and 308 redirects, buffered Content-Length, the empty scalar-field representation, and a known-length file upload alongside a typed-nil reader.

Fresh CI, CodeQL and Castiron runs are currently action_required pending maintainer approval.

markstuart-oai:
@codex review

saioai (approved):
approving. typed-nil readers are handled consistently as empty fields, and the retry, redirect and upload tests pass.

sylvesterkaczmarek:
Thanks for reviewing and confirming the typed-nil retry/redirect/upload behaviour. Appreciate the approval.

saioai:
[saibot]

recommend approving after your final review. before, a missing Go upload reader could crash request creation or lose retries. now it becomes an empty field while real files still stream. this small helper fix is justified, and fresh head/base checks support it. No blocking or substantive issues found. conditional approve after the merge conflict and outstanding review are cleared, with any changed head re-reviewed.

local checks: 4 passed; 1 base comparison failed.

limits: Review and execution cover Linux amd64 only. Native Windows/macOS, the full mock-backed scripts/test suite, FIFO/resource subprocess checks and release/build-artifact checks were not run locally. Repository compilation is not a full-suite pass.…

github checks: 11 reported, 0 failing, 0 pending.

reviewed commit: `9ead33b79a0e`. base: `25664bffc86a`.

sylvesterkaczmarek:
Synced this branch with current upstream `main` in signed merge commit `f265162`. Upstream moved the multipart helpers from `pkg/cmd` to `pkg/custom`, so I relocated the typed-nil transport regression into the helper's current package as part of the merge.

Validation on the refreshed tree: `go test ./internal/apiform ./pkg/custom -count=1` passes; `go test ./pkg/cmd -run '^$' -count=1` compiles the generated command package without requiring the repository's external mock server; `git diff --check` is clean. The branch is conflict-free again.

saioai (approved):

saioai:
[saibot]

before: a missing optional upload reader could crash form encoding or make an empty field lose retries and redirects.

now: missing readers become empty fields consistently, while real uploads keep streaming and accurate lengths.

verdict: hold, this is a useful fix; approve once the existing change request is cleared, since the reader behavior now matches the form’s empty-field contract. resolve the outstanding review before approval.

PR #93, opened 2026-08-18
fix: build CLI releases with patched Go toolchains

jbeckwith-oai:
## Summary
- add a shared `.go-version` file containing the supported Go 1.25 release series
- configure CI and release setup to resolve the latest patched Go 1.25 toolchain while preserving `go 1.25.0` as the minimum supported version
- block artifact builds and releases when a version-pinned `govulncheck` finds reachable vulnerabilities in the shipped CLI; release scans run before write-capable credentials are created

## Why
`actions/setup-go` currently reads the exact `go 1.25.0` directive from `go.mod`, so CI and published binaries use Go 1.25.0 instead of the latest security-patched 1.25 release. A `.go-version` release-series selector plus `check-latest: true` picks up future patch releases automatically without adding update automation.

## Verification
- `GOTOOLCHAIN=go1.25.13 go build ./...`
- `GOTOOLCHAIN=go1.25.13 go test ./internal/...`
- `GOTOOLCHAIN=go1.25.13 go test ./pkg/cmd -run 'Test(ValidateBaseURL|FormatJSON|MTLSClientFlags|NewMTLSHTTPClient|MultipartRequestBody|ExactLengthUpload)'`
- `GOTOOLCHAIN=go1.25.13 go mod verify`
- `GOTOOLCHAIN=go1.25.13 go run golang.org/x/vuln/cmd/govulncheck@v1.7.0 ./cmd/openai` reports `No vulnerabilities found.`
- the same scanner against Go 1.25.0 fails on 25 reachable standard-library vulnerabilities
- Linux and Windows cross-compiled CLI binaries both report `go1.25.13`


.github/workflows/ci.yml
```diff
@@ -45,6 +45,9 @@ jobs:
       - name: Bootstrap
         run: ./scripts/bootstrap
 
+      - name: Check shipped CLI for reachable vulnerabilities
+        run: go run golang.org/x/vuln/cmd/govulncheck@v1.7.0 ./cmd/openai
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 11225c7. CI now runs the shared `scripts/govulncheck`, which installs pinned `govulncheck@v1.7.0` once for the host and scans `linux/amd64`, `darwin/amd64`, and `windows/amd64`. The new `scripts/govulncheck-test` runs first and verifies both complete target coverage and immediate failure when a target is vulnerable. All three real scans reported `No vulnerabilities found`, and the protected `build` check remains fail-closed.

.github/workflows/publish-release.yml
```diff
@@ -56,7 +56,10 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5
       with:
-        go-version-file: "go.mod"
+        go-version-file: .go-version
+        check-latest: true
+    - name: Check shipped CLI for reachable vulnerabilities
+      run: go run golang.org/x/vuln/cmd/govulncheck@v1.7.0 ./cmd/openai
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 11225c7. The release workflow now invokes the same shared `scripts/govulncheck` before either write-capable GitHub App token is created. It installs the pinned scanner for the runner once, then checks the shipped `linux/amd64`, `darwin/amd64`, and `windows/amd64` targets; strict Bash error handling aborts the release on the first failure. Verified all three target scans are clean and all three release-platform CLI builds succeed.

scripts/govulncheck
```diff
@@ -0,0 +1,12 @@
...
+
+set -euo pipefail
+
+cd "$(dirname "$0")/.."
+
+go install golang.org/x/vuln/cmd/govulncheck@v1.7.0
+
+for goos in linux darwin windows; do
+  echo "==> Checking shipped CLI for reachable vulnerabilities ($goos/amd64)"
+  GOOS="$goos" GOARCH=amd64 govulncheck ./cmd/openai
```
HAYDEN-OAI: [P1] Scan every shipped GOOS/GOARCH combination before releasing

.goreleaser.yml also publishes darwin/arm64, linux/386, linux/arm, linux/arm64, windows/386, and windows/arm64, but this loop hard-codes GOARCH=amd64. I reproduced the bypass with the real pinned govulncheck@v1.7.0 and an authentic vulnerable dependency reachable only on 386: GOOS=linux GOARCH=amd64 returned No vulnerabilities found / exit 0, while the shipped GOOS=linux GOARCH=386 target reported three reachable advisories (GO-2022-0572, GO-2022-0569, GO-2022-0463) / exit 3. Both CI and the pre-credential release gate therefore stay green while a vulnerable artifact can be signed and published. Iterate the full GoReleaser target matrix, including relevant architecture variants.

.github/workflows/publish-release.yml
```diff
@@ -56,7 +56,10 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5
       with:
-        go-version-file: "go.mod"
+        go-version-file: .go-version
```
HAYDEN-OAI: [P2] Preserve release tooling when checking out an existing release tag

The preceding Ensure release tag is on main step checks out the user-selected tag, so both .go-version here and ./scripts/govulncheck in the next step are resolved from that tagged commit, not the workflow revision. I verified that the latest existing tag, v1.8.0, passes the unchanged main-ancestor check and contains go.mod, but contains neither new file; the pinned setup action consequently fails with The specified go version file at: .go-version does not exist, and the scanner would also be missing. All 12 existing valid release tags therefore become impossible to republish or recover. Keep the workflow revision's toolchain selector and scanner available across the tag checkout while still scanning the tagged sources.

HAYDEN-OAI (commented):
Found two substantive, reproduced regressions: the vulnerability gate omits six shipped architectures, and release-tag checkout makes every existing valid release tag fail before scanning.

scripts/govulncheck
jbeckwith-oai: Fixed in e37fda0b8fca60e28f28679e978cab7f64b25cbf. The shared, fail-closed scanner now covers the complete GoReleaser release matrix: darwin/amd64, darwin/arm64, linux/386, linux/arm with GOARM=6, linux/amd64, linux/arm64, windows/386, windows/amd64, and windows/arm64. I first verified the strengthened regression test fails against the previous amd64-only scanner, then verified it passes with the complete matrix and stops immediately on a simulated linux/386 vulnerability. An independent workflow check also confirms the scan targets exactly match .goreleaser.yml, and actual govulncheck runs with Go 1.25.13 reported no vulnerabilities for all nine targets.

.github/workflows/publish-release.yml
jbeckwith-oai: Fixed in e37fda0b8fca60e28f28679e978cab7f64b25cbf. The release workflow now resolves the latest configured Go 1.25 patch and preserves the trusted workflow-revision scanner before checking out the selected historical release tag; after checkout, the saved scanner explicitly scans the tagged $GITHUB_WORKSPACE before any release app credentials are minted. The regression test copies the scanner outside a source tree that contains neither .go-version nor scripts/govulncheck and verifies all nine release targets are scanned there. I also reproduced the real release scenario against an archived v1.8.0 tag, confirmed both files are absent, and ran the preserved scanner successfully for all nine targets with Go 1.25.13. The newly merged verified-GoReleaser setup remains intact.

scripts/govulncheck
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+cd "${1:-$(dirname "$0")/..}"
+
+go install golang.org/x/vuln/cmd/govulncheck@v1.7.0
+
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 3eb7da8760eb302b87a2d5845f29bd4a0012f2d7. After installing pinned govulncheck@v1.7.0, the shared scanner now resolves `go env GOBIN` or, when unset, the first `go env GOPATH` entry plus `/bin`, and directly executes that exact installed binary for every release target instead of searching PATH. I added fail-first regression coverage with a hostile shadow executable earlier on PATH (the previous implementation fails with exit 97), plus explicit configured-GOBIN and multi-entry-GOPATH cases. The fixed suite also verifies all nine shipped GOOS/GOARCH targets, Linux ARMv6, fail-closed scanner errors, and historical-tag checkouts. Validation passed with Go 1.25.13: all nine real vulnerability scans on current and v1.8.0 sources, the full mock-backed [go/Windows](https://www.golinks.io/Windows?trackSource=github) test suite, race tests, go vet, fail-closed dependency bootstrap, the verified GoReleaser installer, and a real nine-platform GoReleaser snapshot release.

scripts/govulncheck
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+cd "${1:-$(dirname "$0")/..}"
+
+go install golang.org/x/vuln/cmd/govulncheck@v1.7.0
```
HAYDEN-OAI: [P1] Verify scanner source before installing it from a shared cache

actions/setup-go restores the same main-branch Go module/build cache populated by lower-privilege main CI jobs, including scripts/test -> scripts/mock -> npm exec. On a new or evicted cache key, compromised npm/transitive code can alter the extracted cached golang.org/x/vuln@v1.7.0 before that job saves it for the release workflow. I reproduced this with the original checksum-valid module archive, GOSUMDB=sum.golang.org, and the actual Go 1.25.14 toolchain: go mod verify correctly reports "dir has been modified", but this go install ...@v1.7.0 executes the poisoned extracted source, all nine scans return false-clean/exit 0, and go version -m still claims the legitimate pinned version and h1 checksum. The fake scanner also appended a rogue directory to $GITHUB_PATH; the subsequent goreleaser release --clean then received simulated release-token, Homebrew-token, and macOS-signing environments, so running the gate before credential creation does not contain the compromise. Build the scanner from a fresh, isolated, non-restored module/build cache or explicitly verify the scanner module and dependencies before executing them, and preserve the trusted release executable against post-scan PATH injection.

scripts/govulncheck
```diff
@@ -0,0 +1,23 @@
...
+  scanner_directory="${gopath%%:*}/bin"
+fi
+
+for target in \
+  darwin/amd64 darwin/arm64 \
+  linux/386 linux/arm/6 linux/amd64 linux/arm64 \
+  windows/386 windows/amd64 windows/arm64; do
+  IFS=/ read -r goos goarch goarm <<< "$target"
+  echo "==> Checking shipped CLI for reachable vulnerabilities ($target)"
+  GOOS="$goos" GOARCH="$goarch" GOARM="${goarm:-}" \
```
HAYDEN-OAI: [P1] Match the Linux scan's CGO setting to the shipped binary

.goreleaser.yml forces CGO_ENABLED=0 for every Linux release, but this scan leaves it unset; on the actual Linux/amd64 runner, GOOS=linux GOARCH=amd64 go env CGO_ENABLED is 1. I reproduced a real false-negative with pinned govulncheck@v1.7.0, Go 1.25.14, and golang.org/x/text@v0.3.7 reachable only from a //go:build linux && amd64 && !cgo file: the unchanged nine-platform wrapper reports nine clean scans/exit 0, while scanning the actual release configuration with CGO_ENABLED=0 reports reachable GO-2022-1059 and exits 3. The actual Linux/amd64 binary built with the GoReleaser settings embeds the vulnerable dependency. Set CGO_ENABLED=0 for Linux scans (and otherwise match each shipped target's build environment) so the gate analyzes the code that is signed and published.

HAYDEN-OAI (commented):
Reproduced two P1 security bypasses at the requested head: shared-cache scanner provenance permits a main-CI-to-release credential pivot, and mismatched Linux CGO settings hide an authentic reachable advisory in a shipped binary.

scripts/govulncheck
jbeckwith-oai: Fixed in 0062666e8a8450351e4ca4ec0729a9e1e9ddd9a0 at the scanner/release trust boundary. The pinned govulncheck is now installed and executed from a fresh private GOMODCACHE, GOCACHE, and GOBIN; public module verification is forced with GOSUMDB=sum.golang.org and empty GONOSUMDB/GOPRIVATE, and GOFLAGS=-modcacherw guarantees the private workspace can be removed cleanly. The privileged publish job also disables shared Go cache restoration entirely. Finally, the checksum-verified GoReleaser action exposes its absolute executable path as an action output, and both CI and publishing invoke that exact file instead of consulting PATH. Added a fail-first regression that poisons inherited shared module/build/bin caches, disables sum verification, installs a rogue scanner that modifies GITHUB_PATH, and places another hostile scanner first on PATH; the old implementation fails and the new implementation passes. Verified with Go 1.25.14: nine current-source scans plus nine real historical-release scans all report no vulnerabilities, and a complete nine-platform GoReleaser snapshot succeeds with a malicious goreleaser first on PATH.

scripts/govulncheck
jbeckwith-oai: Fixed in 0062666e8a8450351e4ca4ec0729a9e1e9ddd9a0 by making vulnerability analysis follow the shipped GoReleaser build matrix rather than the runner defaults: every Linux target is scanned with CGO_ENABLED=0, including linux/386, linux/arm/v6, linux/amd64, and linux/arm64; macOS and Windows retain their normal build settings. Added a fail-first regression that records GOOS, GOARCH, GOARM, and CGO_ENABLED for all nine release targets, verifies exact parity with the expected matrix, confirms scan failures stop before later targets, and checks preserved historical-release scanning. The old code fails the Linux-CGO assertions; the fix passes. Verified on the actual Go 1.25.14 toolchain with all nine current-source scans and all nine historical-release scans reporting no vulnerabilities, a complete successful nine-target GoReleaser snapshot, explicit Linux CGO-disabled cross-builds, full CLI tests/Windows test compilation, lint/module verification, and race-tested internal packages.

HAYDEN-OAI (approved):
Verified patched Go 1.25 builds remain compatible with the Go 1.25.0 minimum and preserve the CLI version; confirmed authenticated scanner/module provenance, isolated caches and fail-closed vulnerability handling, all nine shipped platform/architecture targets with Linux CGO parity, historical-tag republication, verified executable paths, and release-token/signing separation. Previously reported issues are fixed; no substantive findings at this commit.

PR #95, opened 2026-08-18
fix: lock and verify Steady mock tooling

jbeckwith-oai:
## Summary

- Replace dynamically executed, Elastic-licensed `@stdy/cli@0.22.1` with reviewed, MIT-licensed Steady `0.22.2`.
- Commit the exact npm manifest and lockfile for the wrapper and five supported native packages, including their reviewed SHA-512 archive integrity values.
- Install the current platform reproducibly with `npm ci --package-lock=true --include=dev --include=optional --ignore-scripts` in a fresh workspace seeded with the reviewed manifests and trusted repository `.npmrc`; never vendor native binaries.
- Use normal Node package resolution to locate the wrapper and native package, require their exact locked names/versions, and constrain their real paths to the validated installation's npm-managed `node_modules` directory.
- Support npm's hoisted, nested, shallow, and linked installation strategies and configured registry mirrors without relying on npm's private installed lockfile.
- Add contributor documentation and focused CI coverage for the actual installation and execution boundaries.

Addresses the P1 Steady/npm test-tooling finding in [SDK-310](https://linear.app/openai/issue/SDK-310/audit-prune-and-vendor-high-risk-dependencies-in-openai-cli).

## Trust boundary

The repository-reviewed `package-lock.json` is the trust root. Every invocation creates an empty installation workspace containing copies of the reviewed `package.json` and `package-lock.json` plus the repository's trusted `.npmrc`, when present; prior forged installations cannot survive dry-run mode, an unreviewed sibling `npm-shrinkwrap.json` cannot override the selected lock, and project-scoped registry/authentication settings remain intact. Explicit `--package-lock=true` prevents npm 7 from silently disabling integrity through inherited configuration. npm verifies downloaded archives against the exact reviewed SHA-512 entries, including when an authorized registry mirror supplies the same bytes. Lifecycle scripts are disabled. The fresh installation is validated before it is published: normal Node package resolution must find the exact-version wrapper/native packages inside its npm-managed `node_modules`, and the executable's real path must remain inside its package. Missing optional natives, `NODE_PATH`/global fallbacks, ancestor fallbacks, incorrect versions, and external symlinks fail closed.

Earlier revisions attempted synthetic provenance verification and inspected npm's private installed lockfile and registry URLs. Those implementation-specific layers could not establish the proposed provenance invariant and rejected supported npm layouts and registry configuration. They have been removed. Publisher provenance remains dependency-update review evidence when available; mock-server startup does not claim independent publisher-provenance verification.

## Validation

- `node scripts/mock-tools/security.test.js` — 17 focused cases cover the six exact MIT-licensed/SHA-512-locked packages, all four installation layouts, missing optional natives, `NODE_PATH`/ancestor fallbacks, incorrect versions, external symlinks, real-npm dry-run reuse, counterfeit same-version shrinkwrap archives, forced npm lockfile enforcement, and repository-only registry settings. All cases pass under npm 7, 10, and 11.
- A counterfeit same-version registry archive executes a synthetic canary under npm 7 with inherited `package-lock=false`, but the exact same archive fails with `EINTEGRITY` when `--package-lock=true` is present.
- Nine end-to-end clean-room installs start real Steady mock servers using npm 7, 10, and 11 with hostile `NODE_ENV=production`, `npm_config_omit=optional`, `npm_config_package_lock=false`, omitted private-lock URLs, an unreachable user-level default registry, and a repository-only configured mirror; npm 10 and 11 cover hoisted, nested, shallow, and linked strategies.
- A local configured registry mirror installs the reviewed bytes even when npm omits installed-lock `resolved` values. A mirror serving a valid but incorrect native archive leaves the optional package unavailable, and the resolver refuses execution.
- `TEST_API_BASE_URL=http://127.0.0.1:14010 ./scripts/test -count=1 -skip TestFilesCreateCLICancelClosesStalledFIFO` — uncached mock-backed Go integration suite and Windows test cross-compilation using an isolated, linked-install Steady 0.22.2 server.
- `./scripts/test-goreleaser-installer`, `go test ./scripts`, `./scripts/lint`, and shell syntax checks pass.
- The unrelated existing `TestFilesCreateCLICancelClosesStalledFIFO` fails independently on this macOS baseline, so only the local invocation excludes it; Linux CI runs the complete unfiltered suite.


scripts/mock
```diff
@@ -23,12 +23,29 @@ fi
...
-  # Pre-install the package so the download doesn't eat into the startup timeout
-  npm exec --package=@stdy/cli@0.22.1 -- steady --version
+# Install the reviewed, integrity-locked native package without lifecycle scripts.
+npm ci --prefix scripts/mock-tools --include=dev --include=optional \
+  --ignore-scripts --no-audit --no-fund
+STEADY="scripts/mock-tools/node_modules/.bin/steady"
+
+# npm registry mirrors do not always expose provenance verification endpoints.
+if [ "${GITHUB_ACTIONS:-}" == "true" ]; then
+  npm audit signatures --prefix scripts/mock-tools
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in ec815ad8c6ffffb016021f226818d8594a59937c. Both the GitHub Actions command and the contributor-documented command now pass `--include=dev --include=optional` to `npm audit signatures`. I reproduced this fail-first with npm’s actual `VerifySignatures#getValidPackageInfo` implementation under `NODE_ENV=production npm_config_omit=optional`: before the change, the audit scope contained only `@stdy/cli`; afterward, it contains both `@stdy/cli` and the installed `@stdy/cli-darwin-arm64@0.22.2` native package. The focused mock-backed integration test and Windows cross-compilation also pass under the same hostile settings.

scripts/test
```diff
@@ -43,7 +43,7 @@ elif ! steady_is_running ; then
   echo -e "To run the server, pass in the path or url of your OpenAPI"
   echo -e "spec to the steady command:"
   echo
-  echo -e "  \$ ${YELLOW}npm exec --package=@stdy/cli@0.22.1 -- steady path/to/your.openapi.yml --host 127.0.0.1 -p 4010 --validator-query-array-format=brackets --validator-form-array-format=brackets --validator-query-object-format=brackets --validator-form-object-format=brackets${NC}"
+  echo -e "  \$ ${YELLOW}./scripts/mock path/to/your.openapi.yml${NC}"
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in ec815ad8c6ffffb016021f226818d8594a59937c. The troubleshooting text now tells contributors to pass their OpenAPI specification directly to `./scripts/mock`, matching the command shown immediately below and the repository-controlled mock-server workflow. Verified with `bash -n scripts/mock scripts/test` and the focused `./scripts/test -run '^TestResponsesCreate$'` run.

CONTRIBUTING.md
```diff
@@ -0,0 +1,165 @@
+# Contributing to the OpenAI CLI
+
+## Setting up the environment
+
+Install [Go 1.25 or later](https://go.dev/doc/install). Running the integration
+tests also requires Node.js, npm, `curl`, and `lsof`.
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in ec815ad8c6ffffb016021f226818d8594a59937c. Reworded the sentence to “The integration tests also require Node.js, npm, `curl`, and `lsof`,” making the plural subject and verb agree while preserving the prerequisite list. `git diff --check` passes.

HAYDEN-OAI (commented):
Validated three supply-chain integrity failures against the exact PR head; the optional-native execution bypass was also reproduced against the base commit. Reproduction details and concrete remediation are in the inline comments.

scripts/mock
```diff
@@ -23,12 +23,29 @@ fi
 
 echo "==> Starting mock server with URL ${URL}"
 
-# Run steady mock on the given spec
-if [ "$1" == "--daemon" ]; then
-  # Pre-install the package so the download doesn't eat into the startup timeout
-  npm exec --package=@stdy/cli@0.22.1 -- steady --version
+# Install the reviewed, integrity-locked native package without lifecycle scripts.
+npm ci --prefix scripts/mock-tools --include=dev --include=optional \
```
HAYDEN-OAI: **[P1] Fail closed when the locked platform-native package is missing**

`--include=optional` does not make the native package mandatory. Serving the exact locked wrapper while returning 404 for `@stdy/cli-linux-x64` made this `npm ci` succeed (`added 1 package`); the Actions signature audit also exited 0 (`audited 1 package`). Steady's `require.resolve` fallback then located and executed an unpinned native binary from an ancestor repository `node_modules`. Under the same registry and ancestor setup, the base commit's cache-isolated `npm exec` failed instead of executing that binary. Require the exact current-OS/architecture native package to exist inside `scripts/mock-tools/node_modules`, validate its locked version and path, and fail closed or invoke that checked native executable directly.

scripts/mock
```diff
@@ -23,12 +23,29 @@ fi
...
+  --ignore-scripts --no-audit --no-fund
+STEADY="scripts/mock-tools/node_modules/.bin/steady"
+
+# npm registry mirrors do not always expose provenance verification endpoints.
+if [ "${GITHUB_ACTIONS:-}" == "true" ]; then
```
HAYDEN-OAI: **[P1] Verify signed provenance for every locked native platform**

This guard only runs in GitHub Actions, whose workflow uses Ubuntu x64. npm's signature audit traverses installed packages, not all lockfile entries, so the exact command successfully audits only 2 of the 6 locked artifacts: `@stdy/cli` and `@stdy/cli-linux-x64`. The `darwin-arm64`, `darwin-x64`, `linux-arm64`, and `win32-x64` native packages are skipped by npm's OS/CPU selectors, and their eventual execution outside Actions skips the audit altogether. Consequently four executable artifacts can lack valid signatures/provenance without any verification gate noticing. Verify all six lockfile artifacts directly, or run fail-closed provenance/signature verification on a matrix covering every supported native platform.

scripts/mock
```diff
@@ -23,12 +23,29 @@ fi
...
+  npm audit signatures --prefix scripts/mock-tools --include=dev --include=optional
```
HAYDEN-OAI: **[P1] Require provenance instead of only checking it when present**

`npm audit signatures` does not require `dist.attestations`: npm verifies provenance only if that metadata field exists, while registry signatures sign `name@version:integrity` and do not protect the field. With the exact locked wrapper and Linux native package, the genuine registry reported two verified attestations; a forwarding registry that removed only `dist.attestations` made this exact audit exit 0 with `2 packages have verified registry signatures` and zero verified attestations. A mirror can therefore strip all SLSA provenance and this new guard still executes Steady. Explicitly fetch and cryptographically verify provenance for every expected executable package, failing when an attestation is absent.

scripts/mock
jbeckwith-oai: Fixed in f41a414d6e4810cbde2729e7f3c4f1b3806c45ad. `scripts/mock-tools/resolve-native.js` derives the exact `@stdy/cli-${platform}-${arch}` package from the checked-in lockfile, verifies its OS/CPU selectors, installed package name and exact version, canonical in-directory package/executable paths, and executable permissions, then `scripts/mock` invokes that checked native executable directly instead of the wrapper’s ancestor-search fallback. Added regression coverage proving a missing native package fails even when an ancestor `node_modules/@stdy/...` exists, mismatched package versions are rejected, and executable symlink escapes fail closed. The real mock-backed integration test and Windows cross-compilation pass with `NODE_ENV=production npm_config_omit=optional`.

scripts/mock
jbeckwith-oai: Fixed in f41a414d6e4810cbde2729e7f3c4f1b3806c45ad. `scripts/mock-tools/verify-provenance.js` builds an isolated metadata-only npm audit project directly from all six committed lockfile entries, so one Ubuntu CI host cryptographically verifies `@stdy/cli` plus `darwin-arm64`, `darwin-x64`, `linux-arm64`, `linux-x64`, and `win32-x64` without downloading or vendoring the other native binaries. Every verified SLSA statement is checked against its exact lockfile SHA-512 and the expected `dgellow/steady` repository/release tag. I validated all six genuine registry signatures and Sigstore bundles through a forwarding registry, and added CI regression tests proving omitted platform artifacts are rejected.

scripts/mock
jbeckwith-oai: Fixed in f41a414d6e4810cbde2729e7f3c4f1b3806c45ad. Verification now runs `npm audit signatures --json --include-attestations` over all six locked artifacts and explicitly requires a cryptographically verified `https://slsa.dev/provenance/v1` bundle for each package; missing/invalid signatures, absent provenance metadata, missing bundles, mismatched signed SHA-512 subjects, or the wrong upstream repository/tag all fail closed. I reproduced the exact forwarding-registry attack: the old command exited 0 with “2 packages have verified registry signatures” after `dist.attestations` was stripped, while the new verifier exited 1 with “Expected verified provenance for all 6 locked Steady packages”; `./scripts/mock` also refused to start against that registry. Dedicated CI regression cases cover both stripped metadata and missing signed bundles.

scripts/mock
jbeckwith-oai: Follow-up in e9a0de2: verified the same metadata-only, six-artifact audit under the Ubuntu runner version npm 10.9.8 and npm 11. Both now must report six cryptographically verified registry signatures and six verified attestations; every SLSA statement is checked against its exact locked SHA-512 plus the dgellow/steady release repository and tag. The full mock-server workflow also passed in CI mode with NODE_ENV=production and npm_config_omit=optional, including Windows cross-compilation.

scripts/mock
jbeckwith-oai: Follow-up in e9a0de2: npm 10 does not expose verified bundles in JSON reports, so the shared npm 10/11 path now requires the full six-signature and six-attestation cryptographic audit summary, then checks each signed SLSA bundle against the lockfile and expected release. Re-ran the adversarial registry that strips only dist.attestations against npm 10.9.8 and npm 11: both fail closed with Expected verified provenance for all 6 locked Steady packages. Added regression coverage for missing or incomplete provenance and incomplete signature coverage.

scripts/mock-tools/verify-provenance.js
```diff
@@ -0,0 +1,133 @@
...
+
+function verifyAuditReport(report, packages) {
+  if (!Array.isArray(report.invalid) || report.invalid.length > 0) {
+    throw new Error("npm reported invalid package signatures or attestations");
+  }
+  if (!Array.isArray(report.missing) || report.missing.length > 0) {
+    throw new Error("npm reported missing package signatures");
+  }
+  if (!Array.isArray(report.verified) || report.verified.length !== packages.length) {
+    throw new Error(`Expected verified provenance for all ${packages.length} locked Steady packages`);
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in e9a0de2, included in the current head: removed the unsupported --include-attestations flag and stopped assuming npm JSON exposes verified or attestationBundles. The verifier now requires the actual npm 10/11 audit summary to report all six verified registry signatures and all six verified attestations, then fetches each signed SLSA bundle and checks its locked SHA-512 plus expected upstream release. Confirmed with the actual runner version npm 10.9.8 and npm 11, including the complete CI-mode mock-server workflow and Windows cross-compilation; an attestation-stripping registry fails closed under both versions.

CONTRIBUTING.md
```diff
@@ -0,0 +1,178 @@
+# Contributing to the OpenAI CLI
+
+## Setting up the environment
+
+Install [Go 1.25 or later](https://go.dev/doc/install). The integration tests
+also require Node.js 14 or later, npm, `curl`, and `lsof`.
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 0b5733d: CONTRIBUTING.md now explicitly requires Node.js 14.18 or later, matching when CommonJS require supports node:-prefixed built-in modules used by scripts/mock and its security helpers. Re-ran all 16 mock-tooling security regressions after updating the documented requirement.

CONTRIBUTING.md
```diff
@@ -0,0 +1,178 @@
+# Contributing to the OpenAI CLI
+
+## Setting up the environment
+
+Install [Go 1.25 or later](https://go.dev/doc/install). The integration tests
+also require Node.js 14.18 or later, npm, `curl`, and `lsof`.
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 7494ddd: CONTRIBUTING now requires npm 7 or later alongside Node.js 14.18 for the lockfileVersion 3 installation path, and separately documents npm 10 or later for the optional signed-provenance verifier. Fail-first proof: npm 6.14.18 rejects the exact checked-in lockfile, while npm 7.24.2 successfully runs npm ci --include=dev --include=optional --ignore-scripts and installs the locked wrapper plus native package.

scripts/mock-tools/verify-provenance.js
```diff
@@ -0,0 +1,183 @@
...
+  if (configuration.error) throw configuration.error;
+  if (configuration.status !== 0) {
+    throw new Error(`Unable to determine the npm registry: ${configuration.stderr}`);
+  }
+
+  const configuredRegistry = configuration.stdout.trim();
+  const registry = configuredRegistry.endsWith("/") ? configuredRegistry : `${configuredRegistry}/`;
+  const verified = packages.map(({ name, version }) => {
+    const endpoint = new URL(`-/npm/v1/attestations/${encodeURIComponent(name)}@${version}`, registry);
+    const response = spawnSync("curl", ["--fail", "--silent", "--show-error", endpoint.href], {
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 7494ddd: replaced unscoped npm config get registry plus bare curl with a small npm-exec helper that loads the actual npm configuration, selects each package registry with npm-registry-fetch.pickRegistry, reads the metadata-advertised attestation URL through pacote, and fetches bundles with npm-registry-fetch so scoped credentials, TLS, proxies, and registry settings are preserved. Fail-first proof: an @stdy-scoped registry requiring Bearer authentication and advertising /custom-attestations failed at the old unscoped registry; the same real six-artifact verification now succeeds with npm 10.9.8 and npm 11. Added a custom advertised-path regression and ran the full CI-mode mock/[go/Windows](https://www.golinks.io/Windows?trackSource=github) workflow against that authenticated registry.

scripts/mock-tools/verify-provenance.js
```diff
@@ -0,0 +1,183 @@
...
+      ({ predicateType }) => predicateType === PROVENANCE_PREDICATE,
+    );
+    if (!provenance?.bundle?.dsseEnvelope?.payload) {
+      throw new Error(`Missing cryptographically verified SLSA bundle for ${expected.name}@${expected.version}`);
+    }
+
+    const statement = JSON.parse(Buffer.from(provenance.bundle.dsseEnvelope.payload, "base64").toString("utf8"));
+    const expectedDigest = Buffer.from(expected.integrity.slice("sha512-".length), "base64").toString("hex");
+    if (statement.subject?.[0]?.digest?.sha512 !== expectedDigest) {
+      throw new Error(`Verified provenance does not match locked SHA-512 integrity for ${expected.name}`);
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 7494ddd: the decoded, cryptographically signed DSSE payload must now contain _type=https://in-toto.io/Statement/v1 and predicateType=https://slsa.dev/provenance/v1 before digest or workflow claims are trusted; unsigned outer labels alone are insufficient. Updated accepted fixtures to include both signed fields and added two fail-first regressions: a missing signed in-toto statement type and a mismatched signed SLSA predicate previously passed but now fail closed. All 19 mock-tooling security regressions pass, and authentic signed bundles for all six locked artifacts verify under npm 10 and npm 11.

scripts/mock-tools/fetch-attestations.js
```diff
@@ -0,0 +1,48 @@
...
+    const manifest = await pacote.manifest(spec, { ...options, fullMetadata: true, preferOnline: true });
+    const attestations = manifest.dist?.attestations;
+    if (!attestations?.url) {
+      throw new Error(`Missing advertised provenance for ${name}@${version}`);
+    }
+
+    const registry = fetch.pickRegistry(spec, options);
+    const response = await fetch(attestationEndpoint(registry, attestations.url), { ...options, registry, spec });
+    const { attestations: attestationBundles } = await response.json();
+    return { name, version, attestations, attestationBundles };
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed architecturally in f49f717: deleted fetch-attestations.js, the independent second registry request, all custom DSSE decoding, synthetic bundle reports, and unsupported source/tag claims. The synthetic all-platform workspace now invokes npm audit signatures exactly once; npm is the sole component that fetches and cryptographically verifies the registry signatures and attestations, and we fail closed unless its exit succeeds and its npm-10/11-compatible summary reports complete coverage for all six locked packages. A fail-first regression proved the previous implementation performed an unexpected second npm invocation; the new test now asserts exactly one audit signatures invocation and rejects both nonzero audit exits and missing/incomplete signature or attestation counts. Verified npm 7 locked installation, authentic six-package scoped/authenticated audits on npm 10 and 11, provenance-stripping rejection on both, the full CI-mode mock/[go/Windows](https://www.golinks.io/Windows?trackSource=github) workflow, GoReleaser tamper tests, and lint. This removes 182 net lines rather than adding another unverifiable cryptographic layer.

scripts/mock-tools/verify-provenance.js
```diff
@@ -0,0 +1,105 @@
...
+    path.join(workspace, "package.json"),
+    JSON.stringify({ name: "openai-cli-steady-provenance-audit", private: true, dependencies }),
+  );
+
+  // npm verifies registry signatures and attestations from package metadata;
+  // manifests let one host audit every platform without downloading its binary.
+  for (const { name, version } of packages) {
+    const directory = path.join(workspace, "node_modules", name);
+    fs.mkdirSync(directory, { recursive: true });
+    fs.writeFileSync(path.join(directory, "package.json"), JSON.stringify({ name, version }));
```
HAYDEN-OAI: **[P1] Bind audited provenance to each locked SHA-512**

The synthesized installed manifest here contains only `name`/`version`, so `npm audit signatures` verifies registry metadata for that version without comparing its cryptographically verified digest with the checked-in lock. I replaced the Linux-x64 lock entry's `resolved` URL and SHA-512 with a hostile same-version `@stdy/cli-linux-x64@0.22.2` tarball: real `npm ci --ignore-scripts` installed it, the unchanged verifier using a fresh cache and the genuine public registry passed signatures/attestations for all six packages, and the resolved native executable read a synthetic `OPENAI_API_KEY` canary. Every checked-in security regression also passes with the locked native SHA-512 replaced by all zeroes. Require each npm-verified SLSA subject digest to equal its corresponding committed lock integrity before the native binary can run; name/version and aggregate audit counts are insufficient.

scripts/mock-tools/verify-provenance.js
```diff
@@ -0,0 +1,105 @@
...
+
+    if (audit.error) throw audit.error;
+    if (audit.status !== 0) {
+      throw new Error(`npm signature and provenance verification failed:\n${audit.stderr}${audit.stdout}`);
+    }
+
+    // npm 10 and 11 omit attestation coverage from their JSON audit reports.
+    // Their shared summary and successful exit are the compatibility boundary;
+    // npm alone fetches and cryptographically verifies every attestation.
+    verifyAuditOutput(audit.stdout, packages.length);
```
HAYDEN-OAI: **[P1] Reject absent or non-SLSA attestation bundles**

This trusts npm's `6 packages have verified attestations` summary as proof that six SLSA provenance bundles were verified, but npm increments that count when package metadata advertises attestations even if the endpoint returns none. A transparent localhost proxy preserved the genuine public npm packuments, signing keys, registry signatures, lockfile, and versions while replacing all six attestation responses with `{"attestations":[]}`: the unchanged verifier exited 0 and claimed all six packages were verified after 12 genuine bundles were removed and zero cryptographic attestations remained. Retaining only the genuinely signed npm publish bundles while removing every SLSA bundle also passes. Require an npm version/API exposing npm-verified bundles and require a cryptographically verified `https://slsa.dev/provenance/v1` bundle for each locked package; npm 10's text summary cannot establish this invariant.

HAYDEN-OAI (commented):
Independently re-reviewed exact head `f49f717b95a0c30dee6a5ff6765eced055bb7ddc`: two reproducible P1 supply-chain verification bypasses remain. Both findings include real npm and registry reproductions inline.

scripts/mock-tools/verify-provenance.js
jbeckwith-oai: Addressed by changing the security boundary in 6917637 rather than claiming an audit can validate a digest it never receives. The entire synthetic provenance workspace and npm audit verifier have been deleted. The reviewed committed package-lock.json is now the explicit trust root: npm ci --ignore-scripts verifies downloaded wrapper/current-platform archives against their committed SHA-512 values, and resolve-native.js additionally compares npm generated node_modules/.package-lock.json entries for both packages against the committed version, resolved registry URL, and SHA-512 before resolving the exact executable. Added fail-first regressions for independently tampered installed native and wrapper digests, a different installed registry URL, and malformed committed SHA-512; all now fail closed, alongside wrong version, ancestor fallback, and escaping executable checks. The PR description and contributor guide explicitly document that committed lockfile changes themselves require dependency-update review rather than pretending publisher provenance authenticates an attacker-modified lock.

scripts/mock-tools/verify-provenance.js
jbeckwith-oai: Addressed architecturally in 6917637 by removing npm audit signatures, the synthetic all-platform workspace, aggregate attestation-count parsing, and every executable runtime claim about verified SLSA bundles. The PR now makes only the supportable invariant explicit: review the exact MIT-licensed Steady 0.22.2 lockfile; npm ci --ignore-scripts enforces its archive SHA-512; compare installed wrapper/native version, registry URL, and integrity to that committed lock; execute only the exact current-platform binary without ancestor fallback. Upstream publisher signatures/provenance are documented solely as evidence to inspect during dependency-update review, not as a runtime gate or guarantee. Nine focused hostile-install tests, real npm 7/10 installation paths, the complete mock-backed [go/Windows](https://www.golinks.io/Windows?trackSource=github) workflow, GoReleaser tamper checks, and lint all pass. Removed the entire 105-line verifier instead of adding another attestation layer.

scripts/mock-tools/resolve-native.js
```diff
@@ -0,0 +1,76 @@
...
+  ]) {
+    if (!/^sha512-[A-Za-z0-9+/]{86}==$/.test(committedPackage.integrity)) {
+      throw new Error(`Locked Steady package ${installedPath} has invalid SHA-512 integrity`);
+    }
+
+    const installed = installedLock.packages?.[installedPath];
+    if (
+      !installed ||
+      installed.version !== committedPackage.version ||
+      installed.resolved !== committedPackage.resolved ||
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 4c4f151 by passing `--omit-lockfile-registry-resolved=false` directly to the canonical `npm ci` install. Before the fix, `npm_config_omit_lockfile_registry_resolved=true` reproduced a successful install followed by native verification failure because the hidden lock omitted `resolved`; the new regression invokes the real `scripts/mock` install under that hostile setting and failed first. After the fix, npm 7, 10, and 11 preserve the exact reviewed wrapper/native URLs, and the committed versions and SHA-512 integrity values continue to validate.

scripts/mock
```diff
@@ -23,12 +23,24 @@ fi
 
 echo "==> Starting mock server with URL ${URL}"
 
-# Run steady mock on the given spec
-if [ "$1" == "--daemon" ]; then
-  # Pre-install the package so the download doesn't eat into the startup timeout
-  npm exec --package=@stdy/cli@0.22.1 -- steady --version
+# Install the reviewed, integrity-locked native package without lifecycle scripts.
+npm ci --prefix scripts/mock-tools --include=dev --include=optional \
+  --ignore-scripts --no-audit --no-fund
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 4c4f151 by passing `--replace-registry-host=never` directly to the canonical `npm ci` install, so npm cannot substitute a configured registry mirror for the exact reviewed npmjs.org tarball URLs before the hidden-lock comparison. The new regression invokes the real `scripts/mock` command with hostile `npm_config_replace_registry_host=always` and failed first until this explicit override was added. npm 7, 10, and 11 were then exercised with hostile registry-host and omission settings; each installed the exact locked wrapper/native pair, retained the committed npmjs.org URLs and SHA-512 integrity, and resolved only the reviewed native executable.

scripts/mock
```diff
@@ -23,12 +23,25 @@ fi
...
 echo "==> Starting mock server with URL ${URL}"
 
-# Run steady mock on the given spec
-if [ "$1" == "--daemon" ]; then
-  # Pre-install the package so the download doesn't eat into the startup timeout
-  npm exec --package=@stdy/cli@0.22.1 -- steady --version
+# Preserve and use the reviewed lockfile URLs without running lifecycle scripts.
+npm ci --prefix scripts/mock-tools --include=dev --include=optional \
+  --ignore-scripts --no-audit --no-fund \
+  --omit-lockfile-registry-resolved=false --replace-registry-host=never
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed architecturally in 9b0d170bc34ea89753cf22ebea10cdab24b47b2d: removed the hoisted-layout assumption and private installed-lock inspection instead of pinning another ambient npm setting. The resolver now uses normal Node resolution from the resolved wrapper, verifies exact locked wrapper/native versions, and constrains both real package paths plus the executable to their trusted boundaries. Real npm 10 and npm 11 installs passed hoisted, nested, shallow, and linked strategies under NODE_ENV=production, npm_config_omit=optional, and omitted private-lock URLs; npm 7 also installed successfully. Focused tests cover all four layouts, missing native packages, ancestor fallback, wrong versions, and external symlinks. A configured mirror installed the locked bytes successfully, while a tampered optional native archive failed closed. The uncached mock-backed suite served 522 valid requests and Windows cross-compilation passed.

HAYDEN-OAI (commented):
Validated three distinct P1 supply-chain execution bypasses against this exact head. Each reproduction used the unchanged launcher, the byte-identical reviewed package-lock, real npm 10, and harmless synthetic credential canaries; concrete failure modes and remediation are in the inline comments.

scripts/mock
```diff
@@ -23,12 +23,24 @@ fi
 
 echo "==> Starting mock server with URL ${URL}"
 
-# Run steady mock on the given spec
-if [ "$1" == "--daemon" ]; then
-  # Pre-install the package so the download doesn't eat into the startup timeout
-  npm exec --package=@stdy/cli@0.22.1 -- steady --version
+# npm verifies every downloaded package against the reviewed lockfile integrity.
+npm ci --prefix scripts/mock-tools --include=dev --include=optional \
```
HAYDEN-OAI: **[P1] Override inherited dry-run mode before trusting `npm ci`**

`npm ci` honors `npm_config_dry_run=true` or an equivalent npmrc setting and exits successfully without deleting or reinstalling existing `node_modules`, so no reviewed archive integrity is enforced. With the unchanged launcher/lock, forged same-version `@stdy/cli` and `@stdy/cli-linux-x64` packages, and `npm_config_dry_run=true npm_config_offline=true`, real npm 10 reported a successful install and the resolver executed the forged native binary, which read a synthetic `OPENAI_API_KEY`. Pass `--no-dry-run`/`--dry-run=false`, or explicitly reject effective dry-run configuration, before treating this command as the integrity boundary.

scripts/mock-tools/resolve-native.js
```diff
@@ -0,0 +1,60 @@
...
+
+function requireInside(directory, candidate, description, boundary = "its locked tooling directory") {
+  const relative = path.relative(directory, candidate);
+  if (!relative || relative === ".." || relative.startsWith(`..${path.sep}`) || path.isAbsolute(relative)) {
+    throw new Error(`${description} resolves outside ${boundary}: ${candidate}`);
+  }
+}
+
+function resolveLockedPackage(toolsDirectory, name, searchDirectory, expectedVersion) {
+  const manifest = require.resolve(`${name}/package.json`, { paths: [searchDirectory] });
```
HAYDEN-OAI, line 12: **[P1] Reject `NODE_PATH` fallback when the locked native package is absent**

`require.resolve(..., { paths })` still consults `NODE_PATH`, while npm treats a failed optional native download as a successful install. I served the genuine SHA-512-verified wrapper but returned 404 for the locked Linux native archive; real npm 10 successfully installed only the wrapper, then `NODE_PATH=scripts/mock-tools/unreviewed` made this resolver select a forged same-name/version native package outside `node_modules`. It survived `npm ci`, passed the broad tooling-directory containment and version checks, and the unchanged launcher executed it with a synthetic `OPENAI_API_KEY`. Exclude `NODE_PATH`/global lookup and require npm-managed installed package locations, including legitimate nested/linked layouts, so missing locked optional natives fail closed.

scripts/mock-tools/resolve-native.js
```diff
@@ -0,0 +1,60 @@
...
+  if (installed.name !== name || installed.version !== expectedVersion) {
+    throw new Error(`Installed Steady package does not match locked ${name}@${expectedVersion}`);
+  }
+
+  return packageDirectory;
+}
+
+function resolveNativeBinary(toolsDirectory, platform = process.platform, architecture = process.arch) {
+  const trustedDirectory = fs.realpathSync(toolsDirectory);
+  const lockfile = JSON.parse(fs.readFileSync(path.join(trustedDirectory, "package-lock.json"), "utf8"));
```
HAYDEN-OAI, line 26: **[P1] Ensure npm installed from this reviewed lock rather than `npm-shrinkwrap.json`**

npm prioritizes a sibling `npm-shrinkwrap.json` over `package-lock.json`, but this resolver always reads the latter and subsequently authenticates installed packages only by name/version. I left the reviewed `package-lock.json` byte-identical, added a shrinkwrap pointing the same-version wrapper/native entries at counterfeit local archives with attacker-selected SHA-512 values, and ran the unchanged launcher with real npm 10; `npm ci --ignore-scripts` installed both counterfeit packages and their native executable read a synthetic `OPENAI_API_KEY`. Reject any overriding `npm-shrinkwrap.json` before installation, or otherwise prove npm's actual selected lock is the reviewed trust root.

scripts/mock
jbeckwith-oai: Fixed architecturally in d7c9bcefdbd3b9c86693277ad4e2bfca66f62144. I first reproduced this fail-first with real npm 10: npm_config_dry_run=true made npm ci report success over forged, same-version wrapper/native packages, and the unchanged launcher executed a synthetic OPENAI_API_KEY canary. The launcher now creates a new unpredictable staging directory containing only copies of the two committed manifests and validates the fresh installation before replacing the published node_modules. Dry-run leaves that new directory without installed packages, so resolution fails before the old forged executable can run. The committed hostile regression now passes under npm 7, 10, and 11; no ad-hoc dry-run flag or npm configuration sandbox is required.

scripts/mock-tools/resolve-native.js
jbeckwith-oai: Fixed in d7c9bcefdbd3b9c86693277ad4e2bfca66f62144. I reproduced the complete exploit fail-first with real npm 10 and a local registry: the genuine locked wrapper installed, the optional native tarball returned 404, NODE_PATH pointed at a forged same-name/version package under the tooling directory, and the old resolver executed its synthetic API-key canary. Package resolution now requires the canonical real path of both wrapper and native packages to remain inside the fresh installation's npm-managed node_modules root, not merely somewhere under the wider tooling directory. NODE_PATH/global fallbacks outside that root are rejected, while actual hoisted, nested, shallow, and linked npm 10/11 installations all start Steady successfully. The dedicated NODE_PATH regression passes under npm 7, 10, and 11.

scripts/mock-tools/resolve-native.js
jbeckwith-oai: Fixed architecturally in d7c9bcefdbd3b9c86693277ad4e2bfca66f62144. I reproduced this fail-first by leaving the reviewed package-lock byte-identical, generating counterfeit same-version wrapper/native tarballs with attacker-controlled SHA-512 values, and adding a sibling npm-shrinkwrap.json; real npm installed the counterfeit archives and the unchanged launcher executed the synthetic API-key canary. Every install now occurs in a fresh directory seeded exclusively with copies of the committed package.json and package-lock.json, so the sibling shrinkwrap is never present in npm's installation workspace and cannot replace the reviewed lock. The candidate installation is validated before publishing any node_modules. The committed counterfeit-shrinkwrap regression passes under npm 7, 10, and 11, and configured mirrors still work when serving the exact reviewed bytes.

scripts/mock
```diff
@@ -23,12 +23,32 @@ fi
...
-if [ "$1" == "--daemon" ]; then
-  # Pre-install the package so the download doesn't eat into the startup timeout
-  npm exec --package=@stdy/cli@0.22.1 -- steady --version
+# Install only reviewed manifests into a new directory, never over prior packages.
+INSTALL_DIRECTORY="$(mktemp -d scripts/mock-tools/.install.XXXXXXXX)"
+trap 'rm -rf "$INSTALL_DIRECTORY"' EXIT
+cp scripts/mock-tools/package.json scripts/mock-tools/package-lock.json "$INSTALL_DIRECTORY/"
+
+npm ci --prefix "$INSTALL_DIRECTORY" --include=dev --include=optional \
+  --ignore-scripts --no-audit --no-fund
```
HAYDEN-OAI, line 42: **[P1] Force package-lock integrity on supported npm 7**

`npm ci` inherits `package-lock=false` under the documented npm 7 minimum because npm 7 does not force its Arborist lockfile option on. With the unchanged launcher and byte-identical reviewed lockfile, real npm 7.24.2 plus `npm_config_package_lock=false` and a configured loopback registry installed counterfeit same-version `@stdy/cli` and `@stdy/cli-linux-x64` archives using attacker-supplied integrity values, then executed their forged native binary with a synthetic `OPENAI_API_KEY` (exit 0). Identical npm 7 with locking enabled and npm 10 rejected the archives with `EINTEGRITY`. Pass `--package-lock=true` explicitly, reject disabled effective lock configuration, or require an npm version whose `npm ci` enforces the lock before treating this install as the reviewed integrity boundary.

scripts/mock
```diff
@@ -23,12 +23,32 @@ fi
...
-# Run steady mock on the given spec
-if [ "$1" == "--daemon" ]; then
-  # Pre-install the package so the download doesn't eat into the startup timeout
-  npm exec --package=@stdy/cli@0.22.1 -- steady --version
+# Install only reviewed manifests into a new directory, never over prior packages.
+INSTALL_DIRECTORY="$(mktemp -d scripts/mock-tools/.install.XXXXXXXX)"
+trap 'rm -rf "$INSTALL_DIRECTORY"' EXIT
+cp scripts/mock-tools/package.json scripts/mock-tools/package-lock.json "$INSTALL_DIRECTORY/"
+
+npm ci --prefix "$INSTALL_DIRECTORY" --include=dev --include=optional \
```
HAYDEN-OAI: **[P2] Preserve repository-level npm registry configuration**

Passing `--prefix "$INSTALL_DIRECTORY"` makes npm load project configuration only from the empty staging directory, so the checkout's existing `.npmrc` and its project-scoped registry settings are silently ignored. In a clean-room reproduction, a repository-level npmrc pointed to a working loopback mirror serving genuine registry archives while the user-level default registry was unreachable: the previous exact `npm exec --package=@stdy/cli@0.22.1 -- steady --version` succeeded through the project mirror, but this unchanged `npm ci --prefix` never contacted that mirror and failed with `ECONNREFUSED`. This prevents mock-server startup in environments that require repository-scoped mirrors or scoped registry/authentication configuration. Preserve the effective project-scoped npm registry settings explicitly without reintroducing an untrusted shrinkwrap into the isolated installation.

HAYDEN-OAI (commented):
Independently reviewed exact head `d7c9bcefdbd3b9c86693277ad4e2bfca66f62144`: one reproduced P1 npm 7 lockfile-integrity bypass executes a counterfeit native binary, and one reproduced P2 loses repository-scoped registry configuration. Both concrete failure modes and remediation directions are documented inline.

scripts/mock
jbeckwith-oai: Fixed in efdf4a8364eabe3925fae43c6288046e2f31ea9c. I reproduced the complete npm 7.24.2 exploit with a loopback registry serving counterfeit same-version wrapper/native tarballs: npm_config_package_lock=false let npm ci ignore the unchanged reviewed lockfile and execute the forged native binary, which wrote a synthetic API-key canary. The clean-room install now makes its trust contract explicit with npm ci --package-lock=true, which has higher precedence than inherited/project npm settings. Against the exact same counterfeit registry and npm_config_package_lock=false, npm 7 now rejects the wrapper with EINTEGRITY using the committed SHA-512 digest. The new focused regression and all 17 security cases pass under npm 7, 10, and 11; nine real server/layout scenarios also pass with locking disabled in the inherited environment.

scripts/mock
jbeckwith-oai: Fixed in efdf4a8364eabe3925fae43c6288046e2f31ea9c. I reproduced the configuration loss fail-first with a repository .npmrc pointing at a project-only registry and a user-level default pointing at a different unreachable registry: the published staged npm ci selected the user registry instead of the repository registry. The clean-room installer now copies only the trusted root .npmrc into staging alongside the two reviewed manifests, preserving project-level and scoped registry/auth settings without copying npm-shrinkwrap.json or weakening the enforced package lock. The focused regression proves the staged install contacts the repository registry and never the unreachable user registry; nine end-to-end npm 7/10/11 hoisted/nested/shallow/linked scenarios then started real Steady servers using only the repository-scoped mirror.

scripts/mock
```diff
@@ -23,12 +23,35 @@ fi
...
-# Run steady mock on the given spec
-if [ "$1" == "--daemon" ]; then
-  # Pre-install the package so the download doesn't eat into the startup timeout
-  npm exec --package=@stdy/cli@0.22.1 -- steady --version
+# Install only reviewed manifests into a new directory, never over prior packages.
+INSTALL_DIRECTORY="$(mktemp -d scripts/mock-tools/.install.XXXXXXXX)"
+trap 'rm -rf "$INSTALL_DIRECTORY"' EXIT
+cp scripts/mock-tools/package.json scripts/mock-tools/package-lock.json "$INSTALL_DIRECTORY/"
+if [[ -f .npmrc ]]; then
+  cp .npmrc "$INSTALL_DIRECTORY/.npmrc"
```
HAYDEN-OAI, line 38: **[P1] Keep copied registry credentials out of the Git worktree**

This copies the repository's npm authentication config into an unignored checkout-local staging directory. In the documented foreground-server workflow, the EXIT trap does not run until the long-lived native process exits. I excluded the original credential-bearing config with an anchored root ignore, ran the unchanged launcher, and executed ordinary git add while the mock server was running: the original remained ignored, but the staging directory's copied config entered the Git index with the complete synthetic registry token. A killed process can retain that copy even longer. Stage outside the worktree, or ignore the entire staging directory and delete the copied authentication file immediately after npm finishes.

scripts/mock
```diff
@@ -23,12 +23,35 @@ fi
...
+fi
+
+npm ci --prefix "$INSTALL_DIRECTORY" --package-lock=true --include=dev --include=optional \
```
HAYDEN-OAI: **[P2] Preserve global npm configuration when changing the staging prefix**

The explicit prefix also changes npm's default globalconfig to the staging prefix's etc/npmrc, so the install silently stops reading the original global npm configuration. With only a synthetic globally configured registry mirror and an empty user config, both npm 10.9.8 and npm 11.19.0 sent the previous npm exec command to the configured mirror, while this exact staged npm ci instead selected registry.npmjs.org; explicitly supplying the original globalconfig restored the mirror. Globally configured private registries, credentials, proxy settings, and custom CAs can therefore be bypassed or lost even though repository-level npm configuration is copied. Capture and preserve the original effective globalconfig when selecting the staging prefix.

HAYDEN-OAI (commented):
Independently reviewed the exact requested head and reproduced two substantive credential-isolation/configuration regressions using synthetic credentials and isolated offline npm configurations.

scripts/mock
jbeckwith-oai: Fixed in ed65ef5189c6d3c99c1150ee802580f86d4e4c04 by moving the complete npm installation into a freshly created OS temporary directory outside the Git worktree, canonicalizing the path for npm 11/macOS, and deleting the staged project `.npmrc` immediately after `npm ci` completes. Added a fail-first regression that writes an anchored-ignored synthetic registry token, invokes real npm, runs ordinary `git add -A` while the install is in progress, and asserts `git grep --cached` cannot find the token. The old checkout-local staging failed by adding `scripts/mock-tools/.install.*/.npmrc`; the new external boundary passes on npm 7, 10, and 11.

scripts/mock
jbeckwith-oai: Fixed in ed65ef5189c6d3c99c1150ee802580f86d4e4c04 by capturing `npm config get globalconfig` before selecting the isolated install prefix, then passing that original effective file explicitly to `npm ci --globalconfig "$NPM_GLOBAL_CONFIG" --prefix "$INSTALL_DIRECTORY"`. This preserves globally configured registry mirrors, credentials, proxies, and CAs while still honoring copied repository-scoped configuration. Added a fail-first real-npm regression with an empty user config and a synthetic mirror available only through the original global prefix; the old code ignored it, while npm 7, 10, and 11 now all use the expected global mirror. All nine real npm/version/layout mock-server scenarios also pass.

jbeckwith-oai:
**P1 addressed: prevent hostile `TMPDIR` from staging registry credentials in the checkout** ([original finding](https://github.com/openai/openai-cli/pull/95#issuecomment-5361546122)).

Fixed in 2adc5d508ee70deab2bf1b156314e14d196d6a87. `scripts/mock` now resolves both the repository root and requested temporary root to their physical filesystem paths before creating anything, and rejects a temporary root equal to or nested beneath the checkout. The existing external staging, immediate `.npmrc` deletion, locked SHA-512 installation, and preserved npm configuration remain intact.

Fail-first regression: initialize a fixture Git worktree with an anchored-ignored synthetic registry token, point `TMPDIR` directly inside that worktree or through an outside symlink back into it, run real npm plus concurrent `git add -A`, and assert `git grep --cached` never finds the token. Both hostile variants failed on the previous head and pass on npm 7, 10, and 11.

jbeckwith-oai:
**P2 addressed: force local mode for the locked Steady installation** ([original finding](https://github.com/openai/openai-cli/pull/95#issuecomment-5361546122)).

Fixed in 2adc5d508ee70deab2bf1b156314e14d196d6a87. The canonical installer now invokes `npm ci --global=false --globalconfig "$NPM_GLOBAL_CONFIG" --prefix "$INSTALL_DIRECTORY" --package-lock=true ...`, so inherited `global=true` / `npm_config_global=true` cannot convert the reviewed local installation into unsupported global mode while the original global registry/proxy/CA configuration remains preserved.

Fail-first regression: set `npm_config_global=true`, configure an isolated synthetic localhost registry, invoke real npm, and require the local installation to reach that registry instead of failing with `ECIGLOBAL`. The old head reproduced `ECIGLOBAL`; the fix passes on npm 7, 10, and 11. All nine real npm-version/layout/mock-server scenarios also passed with `npm_config_global=true` enabled throughout, alongside the full [go/Windows/Python](https://www.golinks.io/Windows/Python?trackSource=github) suites and zero-findings vulnerability scans across nine release targets.

scripts/mock
```diff
@@ -23,12 +23,43 @@ fi
...
+fi
+INSTALL_DIRECTORY="$(mktemp -d "$TEMP_DIRECTORY/openai-cli-mock.XXXXXXXX")"
+trap 'rm -rf "$INSTALL_DIRECTORY"' EXIT
+cp scripts/mock-tools/package.json scripts/mock-tools/package-lock.json "$INSTALL_DIRECTORY/"
+if [[ -f .npmrc ]]; then
+  cp .npmrc "$INSTALL_DIRECTORY/.npmrc"
+fi
 
-  npm exec --package=@stdy/cli@0.22.1 -- steady --host 127.0.0.1 -p 4010 --validator-query-array-format=brackets --validator-form-array-format=brackets --validator-query-object-format=brackets --validator-form-object-format=brackets "$URL" &> .stdy.log &
+npm ci --global=false --globalconfig "$NPM_GLOBAL_CONFIG" --prefix "$INSTALL_DIRECTORY" --package-lock=true --include=dev --include=optional \
```
HAYDEN-OAI: **[P2] Force npm's installation location as well as its global flag**

If a developer or CI image has the valid npm setting `location=global` in either its user or global npmrc, npm 10 and 11 both derive global mode from `location` after applying `--global=false`, so this `npm ci` exits immediately with `ECIGLOBAL: npm ci does not work for global packages`. The previous `npm exec` path works under the same persisted configuration, so this newly prevents mock-server and integration-workflow startup for those users. Pass `--location=project` alongside `--global=false`; that restores the local install and preserves the copied project registry configuration.

scripts/mock
```diff
@@ -23,12 +23,43 @@ fi
...
+  --ignore-scripts --no-audit --no-fund
+rm -f "$INSTALL_DIRECTORY/.npmrc"
+node scripts/mock-tools/resolve-native.js "$INSTALL_DIRECTORY" >/dev/null
+
+rm -rf scripts/mock-tools/node_modules
```
HAYDEN-OAI: **[P2] Serialize publication of the shared mock installation**

Concurrent `./scripts/test` or `./scripts/mock --daemon` invocations can both reach this publication step before the shared health endpoint is available. One invocation can delete `node_modules` after the other has moved its verified installation but before its resolver runs, causing the latter to abort with `Cannot find module '@stdy/cli/package.json'` even though the other daemon successfully starts. I reproduced this with genuine locked npm installations: both daemon invocations succeeded on the base revision against a shared healthy endpoint, whereas this revision returned exit 1 and exit 0. Keep each invocation's verified installation isolated or serialize replacement and resolution.

HAYDEN-OAI (commented):
Independently reviewed exact head `2adc5d508ee70deab2bf1b156314e14d196d6a87`; reproduced two substantive npm-configuration and concurrent-publication regressions. The previously reported credential-isolation and global-configuration issues are fixed.

scripts/mock
jbeckwith-oai: Fixed in e50a24db6e1f9b84cf314835b90c70b23d8d3578 by making the canonical install explicitly project-local: `npm ci --location=project --global=false --globalconfig "$NPM_GLOBAL_CONFIG" --prefix "$INSTALL_DIRECTORY" --package-lock=true ...`.

Added a fail-first real-npm regression that writes `location=global` into a user npmrc, configures a synthetic project-scoped localhost registry, and requires the install to reach that registry instead of failing with `ECIGLOBAL`. The pre-fix revision reproduced `ECIGLOBAL`; both persisted-global-location and inherited-global-flag cases now pass on npm 7, 10, and 11. All nine real npm/version/layout/mock-server scenarios also pass while hostile global configuration remains enabled.

scripts/mock
jbeckwith-oai: Fixed in e50a24db6e1f9b84cf314835b90c70b23d8d3578 by deleting shared publication entirely instead of adding locks or replacement coordination. Each invocation now runs its exact-version, SHA-512-locked native executable directly from its own verified private external installation: `STEADY="$(node scripts/mock-tools/resolve-native.js "$INSTALL_DIRECTORY")"`. The shared `rm -rf`, `mv`, second shared-directory resolution, and now-obsolete checkout `node_modules` ignore are gone.

Added a fail-first lifecycle regression that rejects any shared installation removal/publication and requires resolution directly from the invocation's private directory. Also exercised three genuine pairs of concurrent locked mock-server invocations against a shared healthy endpoint under persisted global-location/global-mode settings: all six launches succeeded, every server returned Steady 0.22.2, and no packages were published into the checkout. Full npm 7/10/11 layout, [go/Windows/Python](https://www.golinks.io/Windows/Python?trackSource=github), and nine-target vulnerability gates pass.

apcha-oai (approved):
[sdk codex review] Approved after two independent full reviews of `2c1e8bb05cbdcc1c204d2b232a4f28ab6decdb2c`. No new actionable findings across all eight changed files, supporting code, correctness, security boundaries, compatibility, maintainability, tests, CI and all 58 reviews / 26 resolved feedback threads.

Verified the reviewed-lockfile → fresh private npm installation → contained exact-version native-executable boundary, configuration preservation, and removal of shared installation publication. The final implementation does not claim independent runtime publisher-provenance verification.

Immediately before submission: open, non-draft, Justin-authored and MERGEABLE; PR base, actual merge base and live main all `3fe27b9a8f6f6df3d8fec5dee0e28badf5d41fb5`. All 13 contexts are terminal: 12 successful and one expected skip. CI checked out `9eaaf80b12fa796e34e2404f840909015146bf55`, whose parents are this exact head and current base and whose tree matches HEAD; 24 mock-tooling regressions, mock-backed Go suites and Windows cross-compilation passed.

Source analysis and existing hosted CI only; no PR code or tests executed locally. Native Windows/macOS and the full minimum-version/install-layout matrix were not independently run.

PR #106, opened 2026-08-19
fix: redact sensitive response headers in debug logs

jbeckwith-oai:
## Summary

- Apply the existing sensitive-header redaction policy to cloned HTTP responses before debug logging.
- Redact redirect `Location` values while preserving original response headers, response identity, unread bodies, and existing request logging behavior.
- Add unit and real-HTTP regression coverage for repeated cookies, authorization values, built-in/custom API-key headers, redirects, and response immutability.

## Validation

- `go test ./internal/debugmiddleware -count=1 -cover` (95.1% statement coverage)
- `go test -race ./internal/debugmiddleware -count=5`
- `go vet ./...`
- `./scripts/lint`
- Full repository integration suite and Windows test cross-compilation passed using a local test-harness overlay that reuses the freshly built CLI.
- Full repository race suite passed with the same overlay.
- One unrelated existing macOS FIFO cancellation test was excluded after reproducing the same failure against the untouched base commit.

Only synthetic credentials and sanitized fixtures are used.


internal/debugmiddleware/debug_middleware.go
```diff
@@ -18,7 +18,7 @@ type (
 const redactedPlaceholder = "<REDACTED>"
 
 // Headers known to contain sensitive information like an API key. Note that this exclude `Authorization`,
-// which is handled specially in `redactRequest` below.
+// which is handled specially in `redactHeaders` below.
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in c6212773df09f6062c999082576da0be785aaf11: changed the header comment to “this excludes `Authorization`” and the `redactRequest` comment to “the original request is returned unchanged.” Verified with `go test ./internal/debugmiddleware -count=1 -cover` (95.1%), `go test -race ./internal/debugmiddleware -count=5`, `go vet ./...`, `./scripts/lint`, and `git diff --check`; response-header redaction behavior is unchanged.

jbeckwith-oai:
Addressed Copilot submitted review [4975555643](https://github.com/openai/openai-cli/pull/106#pullrequestreview-4975555643) in 2f0684d6f22d64adfe4e9856be038f3fe15f838c: removed the duplicated `the` across the wrapped `redactRequest` comment so it reads “As a small optimization, the original request is returned unchanged.” Verified with `go test ./internal/debugmiddleware -count=1 -cover` (95.1%), `go test -race ./internal/debugmiddleware -count=5`, `go vet ./...`, and `./scripts/lint`. Documentation only; no behavior changed.

jbeckwith-oai:
Addressed Copilot submitted review [4975655882](https://github.com/openai/openai-cli/pull/106#pullrequestreview-4975655882) in da569efcb01c57516d31a484f914cea08fd04f4e. Added a fail-first regression proving `(nil, nil)` previously panicked, then extended the existing middleware early return to `if err != nil || resp == nil`; nil responses now pass through unchanged without logging a response, while ordinary response redaction/logging remains covered. Verified: focused middleware tests (97.6% coverage), 10 race repetitions, full integration suite plus Windows cross-compilation, repository race coverage, `go vet ./...`, and `./scripts/lint`. Broad race runs excluded only independently reproduced pre-existing FIFO and requestflag-test issues.

HAYDEN-OAI (commented):
Two independently reproduced response-logging credential disclosures remain: case-variant sensitive headers and populated chunked-response trailers.

internal/debugmiddleware/debug_middleware.go
```diff
@@ -66,9 +66,37 @@ func (m *RequestLogger) Middleware() Middleware {
...
+
+	redacted := req.Clone(req.Context())
+	redacted.Header = redactedHeaders
+	return redacted, nil
+}
+
+// redactResponse clones the response and redacts sensitive headers for logging.
+func (m *RequestLogger) redactResponse(resp *http.Response) *http.Response {
+	redacted := *resp
+	redacted.Header = m.redactHeaders(resp.Header)
```
HAYDEN-OAI: [P1] Redact case-variant duplicates of sensitive response headers

`http.Header.Values` and `Del` canonicalize only the lookup key, so a custom `http.RoundTripper` can return both `Authorization` and `authorization` (likewise `X-Api-Key`/`x-api-key`, `Set-Cookie`/`set-cookie`, and `Location`/`location`). The canonical entry is redacted, but `httputil.DumpResponse` writes the lowercase duplicate verbatim; I reproduced leaked bearer tokens, API keys, session cookies, and signed redirect URLs through `http.Client.Do` on this commit. Match against the actual header-map keys case-insensitively and redact every variant, including `Location`.

internal/debugmiddleware/debug_middleware.go
```diff
@@ -66,9 +66,37 @@ func (m *RequestLogger) Middleware() Middleware {
...
+	}
+
+	redacted := req.Clone(req.Context())
+	redacted.Header = redactedHeaders
+	return redacted, nil
+}
+
+// redactResponse clones the response and redacts sensitive headers for logging.
+func (m *RequestLogger) redactResponse(resp *http.Response) *http.Response {
+	redacted := *resp
```
HAYDEN-OAI, line 84: [P2] Redact populated response trailers before dumping the clone

The shallow response copy shares `resp.Trailer`, but only `resp.Header` is sanitized. `httputil.DumpResponse(..., false)` still serializes populated trailers when a custom transport returns an empty chunked response, so canonical `Authorization`, `X-Api-Key`, and `Set-Cookie` trailer values are written to debug output in cleartext; I reproduced all three through `http.Client.Do`. Clone and apply the same redaction policy to `resp.Trailer` before dumping.

internal/debugmiddleware/debug_middleware.go
jbeckwith-oai: Fixed in 87f3cb3fa64c73f22586902a8d9705001d447b81. `redactHeaders` now iterates actual cloned map keys with `strings.EqualFold` instead of canonical `Values`/`Del`, preserving each key's casing, duplicate values, and bearer scheme while redacting every authorization/API-key/cookie variant; response `Location` variants use the same case-insensitive policy. Added `RedactsCaseVariantResponseHeadersFromCustomTransport` through real `http.Client.Do` plus `RedactsCaseVariantRequestHeadersWithoutMutatingRequest`; both reproduced cleartext leaks before the fix and now pass while proving caller maps and response bodies remain unchanged. Middleware coverage: 97.7%; 25 race runs, full suite, Windows cross-compilation, lint, and vet pass.

internal/debugmiddleware/debug_middleware.go
jbeckwith-oai: Fixed in 87f3cb3fa64c73f22586902a8d9705001d447b81. `redactResponse` now independently clones and applies the identical case-insensitive response-redaction policy to both `resp.Header` and `resp.Trailer`, including authorization, built-in/custom API keys, repeated cookies, and signed `Location` values; the original response and both maps are unchanged. Added fail-first `RedactsCaseVariantResponseTrailersFromCustomTransport`, using real `http.Client.Do` and an empty chunked response with populated canonical/lowercase trailers; it previously leaked credentials and now proves all variants are redacted while benign trailers and the unread body are preserved. Middleware coverage: 97.7%; 25 race runs, full suite, Windows cross-compilation, lint, and vet pass.

HAYDEN-OAI (approved):
Reviewed exact head 6fe9708764efe1e82bd5ddc85ed42b3c1edfabb9. Confirmed the prior mixed-case response-header and populated-trailer credential disclosures are fixed: every case-variant authorization, API-key, cookie, and Location entry is independently redacted, while original response headers/trailers and unread streaming bodies remain unchanged. Verified custom transports, redirects, and nil/error paths; no substantive correctness or security findings.

PR #107, opened 2026-08-19
fix: restrict downloaded file permissions

jbeckwith-oai:
## Summary

- Create newly downloaded files with owner-only permissions for explicit output paths and server-provided filenames.
- Preserve existing output-file permissions and overwrite behavior, filename collision handling, temporary-file fallback, and traversal protections.
- Cover existing-file modes, permissive and restrictive umasks, encoded traversal, symlink collisions, and concurrent downloads with synthetic fixtures.

## Verification

- Focused download tests under umasks `000`, `022`, and `077`.
- `go test -race ./pkg/cmd -run '^(TestWriteBinaryResponse|TestCreateDownloadFile)$' -count=5`
- All six generated download endpoint integration tests.
- `./scripts/lint`, `go vet ./...`, and formatting checks.
- Full repository and race suites, excluding one pre-existing macOS FIFO cancellation failure independently reproduced on the unchanged base commit.
- Windows amd64/arm64 and Linux amd64/arm64 compilation.

### DIFF
--- a/pkg/cmd/cmdutil.go
+++ b/pkg/cmd/cmdutil.go
@@ -218,7 +218,7 @@ func writeBinaryResponse(response *http.Response, stdout io.Writer, outfile stri
 		}
 		return fmt.Sprintf("Wrote output to: %s", file.Name()), nil
 	default:
-		if err := os.WriteFile(outfile, body, 0644); err != nil {
+		if err := os.WriteFile(outfile, body, 0600); err != nil {
 			return "", err
 		}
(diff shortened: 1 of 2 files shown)


HAYDEN-OAI (approved):
Independent security review of the exact requested commit: no introduced P1/P2 findings. Verified owner-only file creation, preserved overwrite behavior, collision and traversal protections, stdout, failure paths, and concurrent downloads with direct base-versus-HEAD reproductions.

PR #111, opened 2026-08-19
fix: prepare release artifacts before publishing

jbeckwith-oai:
## Summary
- Generate shell completions and man pages in an isolated, read-only preparation job before protected release publishing.
- Bind prepared artifacts to the verified release commit and skip legacy GoReleaser preparation hooks when republishing historical tags.
- Preserve existing archive contents, Linux packages, Homebrew output, signing/notarization configuration, attestations, and current pinned GitHub Actions versions.
- Add workflow, credential-boundary, release-configuration, historical-tag, and artifact-compatibility regression coverage.

## Validation
- `env GOCACHE=/Users/jbeckwith/.cache/go-build ./scripts/test -skip '^TestFilesCreateCLICancelClosesStalledFIFO$'`
- `env GOCACHE=/Users/jbeckwith/.cache/go-build go test -race -parallel=1 ./... -skip '^TestFilesCreateCLICancelClosesStalledFIFO$'`
- `env GOCACHE=/Users/jbeckwith/.cache/go-build go test -race ./scripts -count=1`
- `./scripts/lint`, `go vet ./...`, `actionlint`, `goreleaser check`, and `./scripts/test-goreleaser-installer`
- Full GoReleaser snapshot across all nine Linux/macOS/Windows archive targets, plus Linux packages and the Homebrew cask.

The excluded FIFO test fails independently on the unchanged macOS baseline; Windows test cross-compilation passes.


.github/workflows/publish-release.yml
```diff
@@ -29,36 +64,19 @@ jobs:
...
+        tag: ${{ inputs.tag }}
     - name: Set up Go
       uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
       with:
         go-version-file: "go.mod"
+    - name: Download release support artifacts
+      uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4
+      with:
+        name: release-support-artifacts-${{ steps.release-tag.outputs.commit }}
+        path: .
```
HAYDEN-OAI: [P1] Exclude downloaded artifacts from GoReleaser's clean-tree check

Downloading the support artifact into the repository creates untracked `completions/` and `man/` directories; neither the current checkout nor historical release tags ignore those paths. Pinned GoReleaser 2.15.4 runs `git status --porcelain` before processing release hooks and rejects any dirty worktree for non-snapshot releases, so the new `goreleaser release --clean --skip=before` command aborts before building or publishing every release. The previous hooks generated these files only after that validation. Add both directories to `.git/info/exclude` after checking out the release tag; updating only the current `.gitignore` would still break historical-tag releases.

HAYDEN-OAI (commented):
One release-blocking regression in the artifact preparation flow.

.github/workflows/publish-release.yml
jbeckwith-oai: Fixed in 382cf070. After checking out the requested release tag, the protected job now adds rooted `/completions/` and `/man/` patterns to `$(git rev-parse --git-path info/exclude)` before downloading the prepared artifacts. This keeps non-snapshot release checkouts clean without depending on a historical tag's `.gitignore` or changing tracked files.

I also added a fail-first integration test using the verified GoReleaser binary against a real older tagged repository: it first reproduces the dirty-tree failure, then executes the workflow's exact exclusion command and verifies the real non-snapshot release succeeds, preserves all three shell completions and the man page, and never runs legacy hooks. CI now runs that historical release preflight explicitly.

scripts/generate-release-artifacts
```diff
@@ -0,0 +1,31 @@
...
+)
+
+for credential in "${credentials[@]}"; do
+  if [[ -n "${!credential:-}" ]]; then
+    echo "Release artifacts must be generated without $credential" >&2
+    exit 1
+  fi
+done
+
+mkdir -p completions
```
HAYDEN-OAI, line 27: [P2] Keep locally prepared release artifacts out of Git status

Running the documented local sequence (`./scripts/generate-release-artifacts`, then a non-snapshot `goreleaser release --clean`) on a clean checkout creates untracked `completions/` and `man/` files, and neither directory is ignored. GoReleaser 2.15.4 rejects the dirty worktree before building; previously the `before` hooks created these files only after that validation. The hosted publish job now installs `.git/info/exclude` rules, but the standalone generator and local-release instructions do not, so locally prepared tagged releases remain broken. Add rooted ignore rules for current checkouts or apply the same checkout-local exclusions in the standalone preparation flow.

HAYDEN-OAI (commented):
One actionable local tagged-release compatibility regression remains.

scripts/generate-release-artifacts
jbeckwith-oai: Fixed in da8bcecba658616cc5e6bcfe4326931d8f690360 by adding rooted `/completions/` and `/man/` ignore rules for current local checkouts, while preserving the existing equivalent rooted `.git/info/exclude` rules for historical hosted release tags.

CI now executes a real tagged local fixture using the actual `./scripts/generate-release-artifacts` followed by exact non-snapshot `goreleaser release --clean`. The fail-first regression reproduces the original dirty-tree rejection, verifies all three shell completions and the man page remain in the release archive, and confirms an unrelated `nested/completions/unrelated.txt` still fails GoReleaser’s clean-tree check. The existing real historical-tag preflight, all nine release archives and Homebrew output, the full normal and race suites, Windows compilation, workflow lint, and focused credential-isolation/security checks also pass.

.github/actions/checkout-release-tag/action.yml
```diff
@@ -0,0 +1,46 @@
...
+        esac
+        git check-ref-format "refs/tags/$TAG"
+
+        authenticated_origin="https://x-access-token:${GITHUB_TOKEN}@github.com/${GITHUB_REPOSITORY}.git"
+        remote_tag="$(git ls-remote --exit-code --tags --refs "$authenticated_origin" "refs/tags/$TAG")"
+        remote_ref="$(printf '%s\n' "$remote_tag" | awk '{print $2}')"
+        test "$remote_ref" = "refs/tags/$TAG"
+
+        git fetch --no-tags "$authenticated_origin" +refs/heads/main:refs/remotes/origin/main
+        git fetch --no-tags "$authenticated_origin" "+refs/tags/$TAG:refs/tags/$TAG"
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in e71a5f115e344edc2781180ae1604ab7456eebb4. The shared release-tag checkout now uses a clean `https://github.com/${GITHUB_REPOSITORY}.git` remote and GitHub-scoped `git --config-env=http.https://github.com/.extraheader=RELEASE_GIT_AUTHORIZATION` HTTP authentication. The Basic authorization value exists only in the checkout step’s environment for authenticated `ls-remote` and the two tag/main fetches; it is unset before subsequent local Git commands and is never written to Git configuration.

Added a fail-first integration regression that executes the actual composite-action shell script against a real tagged Git repository using a synthetic token. Before the fix, it reproduced the raw token in a Git failure message. Afterward, successful authenticated tag discovery/main ancestry verification/checkout and a failing authenticated fetch both verify that neither the raw token nor its Base64 representation appears in Git process arguments, command output, or `.git/config`; the fixture also rejects authorization leaking into unrelated local Git commands. Full normal and race suites, Windows compilation, hosted-runner simulation, local and historical real GoReleaser preflights, all nine release archives/Homebrew artifacts, credential-isolation regressions, workflow lint, installer verification, and release configuration validation all pass.

HAYDEN-OAI (approved):
Reviewed the exact current head and found no substantive correctness, security, compatibility, or release-pipeline issues.

PR #117, opened 2026-08-20
fix(jsonview): avoid width underflow in static string rendering

sylvesterkaczmarek:
## Summary

Avoid converting a negative remaining display width into an enormous unsigned truncation limit when rendering static JSON strings in narrow terminals.

## Problem

`formatResult` currently truncates long strings with:

```go
truncate.String(str, uint(width-1))
```

Object and array rendering can pass a non-positive `width` after subtracting the rendered key or numbering width. Converting `width-1` to `uint` then wraps the negative value to a very large number, so the string is effectively not truncated and can spill far beyond the available terminal width.

## Fix

Centralize string-width handling in a small helper that:

- returns no string content when no width is available;
- renders a single ellipsis when exactly one column is available;
- preserves strings that already fit;
- only converts a verified positive truncation width to `uint`.

Normal-width output is unchanged.

## Regression coverage

Added focused tests covering:

- a non-positive string width, which previously retained the full string because of unsigned wraparound;
- the one-column case, which now deterministically renders a single ellipsis.

## Validation

The branch is based directly on current upstream `main` (`ee92673a416c0851a5fe8907a2453db7bd450633`) and contains one signed commit touching only the static JSON renderer and its focused regression test. Full Go test execution is left to repository CI.

## Risk

Low. The behavior change is limited to string values whose available display width is zero or too small to contain the original value.

BEGIN_COMMIT_OVERRIDE
fix(jsonview): avoid width underflow in static string rendering (#117)
END_COMMIT_OVERRIDE

### DIFF
--- a/internal/jsonview/staticdisplay.go
+++ b/internal/jsonview/staticdisplay.go
@@ -35,6 +35,19 @@ func formatJSON(json gjson.Result, width int) string {
 	return formatResult(json, 0, width)
 }
 
+func truncateStringToWidth(str string, width int) string {
+	if width <= 0 {
+		return ""
+	}
+	if lipgloss.Width(str) <= width {
+		return str
+	}
+	if width == 1 {
+		return "…"
+	}
+	return truncate.String(str, uint(width-1)) + "…"
+}
+
 func formatResult(result gjson.Result, indent, width int) string {
 	switch result.Type {
 	case gjson.String:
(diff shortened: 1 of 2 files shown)


markstuart-oai:
@codex review

saioai (approved):
the width fix makes sense. i tested the object and array cases against the previous commit: long strings went past the available width before, and they don’t on this branch. the jsonview race tests passed 5 runs, the internal tests passed, and the repo compiled. vet, module verification, formatting, and diff checks passed too. github checks haven’t reported yet, so i’d wait for those before merging.

sylvesterkaczmarek:
Thanks for the review and approval.

sylvesterkaczmarek:
Thanks for getting this merged and for the careful validation. Appreciate the review.

PR #121, opened 2026-08-20
fix(jsonview): preserve literal keys throughout the interactive explorer

sylvesterkaczmarek:
## Summary

Make every interactive JSON-explorer object lookup use literal JSON member names instead of interpreting those names as GJSON path expressions.

This complements #86, which addresses the same class of bug in the static pretty renderer. The interactive explorer has its own independent lookup paths and remains affected without this change.

Related to #81.

## Problem

The explorer enumerates real object member names with `@keys`, but several rendering paths then feed those names back into `gjson.Result.Get`. `Get` parses its argument as a GJSON path, so a literal key such as `a.b` can resolve the nested path `a -> b` instead of the top-level member named exactly `a.b`.

For example:

```json
{
  "a.b": "literal-value",
  "a": {"b": "nested-value"}
}
```

can show `nested-value` under the `a.b` key even though the underlying JSON stores `literal-value` there.

The bug exists in four interactive paths:

1. ordinary object tables;
2. array-of-object tables;
3. compact object previews;
4. rows appended later by the streaming/lazy-load iterator.

Fixing only one of these leaves the same incorrect value substitution elsewhere in the explorer.

## Fix

Once keys have been enumerated, materialize each object's literal map with `Result.Map()` and index it directly by the key string. No GJSON path parsing is used for an already-known object member name.

The change covers all four interactive paths above while leaving genuine navigation-path construction and user transformations unchanged.

## Regression coverage

Added focused tests proving literal dotted keys remain distinct from nested paths in:

- object-table rendering;
- array-of-object column rendering;
- compact object previews;
- lazy-loaded streamed rows.

The fixtures deliberately contain both `"a.b"` and nested `{"a":{"b":...}}` values so a path-based lookup deterministically returns the wrong value before the fix.

## Validation

The branch is based directly on current upstream `main` (`a7719136b8ed401b0c51a05553a5e4f720150307`) and contains one DCO-signed commit. The production change is 9 additions / 4 deletions in handwritten `internal/jsonview/explorer.go`, plus focused regression coverage. Full repository test execution is left to CI.

## Risk

Low. The explorer already has the literal member names. This change only stops reinterpreting those names as query expressions. Simple keys retain the same values, while keys containing GJSON path syntax now correctly refer to their own object members.

### DIFF
--- a/internal/jsonview/explorer.go
+++ b/internal/jsonview/explorer.go
@@ -177,13 +177,14 @@ func (tv *TableView) loadMoreData(raw bool) tea.Cmd {
 
 		// For array of objects, we need to format according to columns
 		if len(tv.columns) > 1 && result.IsObject() {
+			values := result.Map()
 			newRow = make(table.Row, len(tv.columns))
 			for i, col := range tv.columns {
 				key := col.Title
 				if i < len(tv.columnKeys) {
 					key = tv.columnKeys[i]
 				}
-				newRow[i] = formatValue(result.Get(key), raw)
+				newRow[i] = formatValue(values[key], raw)
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


markstuart-oai:
@codex review

saioai (commented):
closing this as superseded by #143, which is now merged and covers the literal-key fix here, including rows loaded later. thanks for the fix.

sylvesterkaczmarek:
Thanks, makes sense. Glad the literal-key fix landed through #143 with the later-row coverage included.

PR #125, opened 2026-08-21
fix(apiform): encode typed nil readers as empty fields instead of panicking

Hughhhhcoder:
## Problem

`encodeValue` routes any value whose type implements `io.Reader` into `encodeReader` **before** checking whether the value is a typed nil. A typed nil pointer (or nil interface) whose type implements `io.Reader` therefore reaches `encodeReader`, which calls `io.Copy` on a nil receiver and panics.

## Fix

Move the nil check for `reflect.Pointer` and `reflect.Interface` ahead of the `io.Reader` interface check, so typed nil values are encoded as empty fields instead of panicking.

## Test

Added a regression test in `internal/apiform/form_test.go` covering typed nil `io.Reader` values. `go test ./internal/apiform/...` passes.

### DIFF
--- a/internal/apiform/encoder.go
+++ b/internal/apiform/encoder.go
@@ -51,15 +51,27 @@ func (e *encoder) encodeValue(key string, val reflect.Value, writer *multipart.W
 
 	t := val.Type()
 
+	// Encode nil pointers and nil interfaces as empty fields before any other
+	// type detection. A typed nil pointer whose type implements io.Reader must
+	// not be routed into encodeReader, which would invoke Read on a nil
+	// receiver via io.Copy and panic.
+	switch t.Kind() {
+	case reflect.Pointer:
+		if val.IsNil() {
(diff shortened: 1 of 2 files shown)


Hughhhhcoder:
The latest pull-request workflow runs (`CI`, `CodeQL`, and `Castiron custom code`) are completed with `action_required`. Could a maintainer approve the workflows so they can run? This appears to be a workflow-permission gate rather than a code failure; the focused and package tests pass locally.

jbeckwith-oai:
Thanks for the contribution. This duplicates #89 and the earlier PR #90, which already owns the original report. While reviewing both, I found that the shallow nil check still panicked when a typed-nil pointer was wrapped in an `io.Reader`-typed map or slice. I have pushed the interface-unwrapping correction and expanded regression coverage to #90, so I am closing this PR in favor of that one.

PR #128, opened 2026-08-21
chore(api): No customer-visible SDK or public API changes

apcha-oai:
## Summary

This update does not modify any customer-facing SDK files or public API contracts. There are no changes to public symbols, method signatures, parameters, return types, or runtime behavior; no release is required.

## Changes

- No changes to the public SDK surface: no added, removed, or modified public symbols, methods, parameters, or return types
- No changes to CLI behavior or public API contracts
- Only non-customer-facing generation metadata was updated; no release impact


jbeckwith-oai (approved):
Approved after two independent reviews: generation metadata only, unchanged generated content hash, no runtime/public API/workflow/security changes, and passing public checks.

PR #136, opened 2026-08-24
feat(api): Add residency flag and deprecate geography

apcha-oai:
## Summary

Introduce a new optional --residency flag to configure project data residency and mark the existing --geography flag as deprecated in admin project commands. This is backward-compatible: geography remains accepted for compatibility, but help text now directs users to prefer residency.

## Changes

- Add an optional --residency flag to the admin organization project create command; the flag value is submitted as the request body field residency.
- Update help text for the --geography flag to indicate deprecation and direct users to use --residency instead when configuring data residency; the update command help clarifies geography is retained for backward compatibility.
- Guidance warns not to provide both --geography and --residency, but this is documented in help text only (no programmatic enforcement of mutual exclusivity).
- No breaking changes: existing --geography usage continues to work, no inputs were made required, and return types and server-side contracts are unchanged.


jbeckwith-oai (changes requested):
Requesting changes for the existing P2 mutual-exclusion finding https://github.com/openai/openai-cli/pull/136#discussion_r3845653965 and invalid-fixture finding https://github.com/openai/openai-cli/pull/136#discussion_r3845654022: the CLI accepts both --geography and --residency despite its explicit prohibition, and tests assert the invalid combination succeeds. Reject combined flags/body keys, split valid legacy and residency test cases, and encode the invariant in upstream schema/code generation.

jbeckwith-oai (approved):
LGTM.

PR #138, opened 2026-08-25
chore(api): update generation metadata only

apcha-oai:
## Summary

Keep the CLI’s generation metadata aligned with the coordinated SDK update.

The CLI still does not expose Realtime call creation. This PR changes only generation metadata; commands, runtime behavior, and the public API reference are unchanged.


jbeckwith-oai (approved):
Verified the exact diff is limited to generation/config/codegen provenance fields in .castiron.stats.yml. OpenAPI content hashes, CLI commands, runtime behavior, authentication, and generated API surfaces are unchanged; CI/security checks pass. Approved.

PR #140, opened 2026-08-26
feat(api): make function call output call IDs optional

apcha-oai:
## Summary

Update the bundled API schema so `call_id` is optional in stable and beta function call output inputs and responses; input values also accept `null`.


jbeckwith-oai (approved):
Verified this change only updates the transformed stable/beta OpenAPI schemas and Castiron generation metadata; optional response IDs and optional/nullable input IDs match the canonical API contract, with no CLI command/runtime compatibility changes. Looks good.

PR #142, opened 2026-08-27
feat(api): add compute_units to Responses and Chat Completions usage

apcha-oai:
Adds `compute_units` to Responses and Chat Completions usage.


HAYDEN-OAI (approved):
Reviewed `d6e48a06cdf7aee6a556758050303d04bdaa80fa`. The three usage schemas consistently add an optional, nullable, non-negative integer, and the CLI's raw-response/RawJSON output paths preserve the new field without requiring response-model changes here. The two event-enum additions are additive. I also verified the transformed specification's MD5 matches the updated generation metadata.

No substantive issues or additional inline findings. The already-reported description wording nit is non-blocking and can be corrected in the authoritative schema.

jbeckwith-oai (approved):
Verified the schema-only update adds optional nullable nonnegative compute_units consistently to Responses and Chat usage without changing CLI runtime behavior or dependencies; generated metadata and canonical enum synchronization are consistent.

PR #143, opened 2026-08-31
fix(jsonview): preserve literal object keys in pretty output and explorer

### LINKED ISSUE #81: Pretty JSON output resolves literal object keys as GJSON paths

Ultron09:
## Summary

Preserve literal object keys across pretty formatting and the interactive JSON explorer instead of resolving member names containing dots or metacharacters as GJSON query paths.

Fixes #81.

## Problem & Root Cause

The JSON renderer and explorer enumerate literal member names with `@keys` / `gjson.Result.Get("@keys")`, but subsequently fetched each value through `result.Get(key.Str)`, `data.Get(key.Str)`, or `item.Get(key)`.

Because GJSON interprets string arguments to `.Get()` as path expressions (such as evaluating `"a.b"` as a traversal into nested object `a -> b`), keys with dots or other GJSON path syntax either:
1. Display a nested child value under a top-level key's label, or
2. Fail to find any value and display empty/null even when the literal key exists.

## Fix

Materialize the object map once with `result.Map()` and look up values by exact literal key:
- `internal/jsonview/staticdisplay.go`: `formatJSONObject`
- `internal/jsonview/explorer.go`: `newObjectTableView`, `newArrayOfObjectsTableView`, `TableView.loadMoreData`, and `formatObject`

## Regression Coverage

Added comprehensive regression tests in `internal/jsonview/literal_keys_test.go`:
- `TestStaticDisplayUsesLiteralObjectKeys`: confirms static pretty formatting preserves top-level dotted keys and nested paths concurrently.
- `TestExploreTableUsesLiteralObjectKeys`: confirms object table view and array-of-objects table view display the literal key's value.
- `TestExploreFormatObjectUsesLiteralObjectKeys`: confirms object previews format literal keys correctly.

## Validation Commands & Results

- `go test -v ./internal/jsonview/...` (PASS)
- `go test ./pkg/cmd -run TestFormatJSON` (PASS)
- `go vet ./internal/... ./pkg/...` (PASS, clean)
- `go mod verify` (PASS, all modules verified)
- `git diff --check` (PASS, clean)

### DIFF
--- a/internal/jsonview/explorer.go
+++ b/internal/jsonview/explorer.go
@@ -183,12 +183,13 @@ func (tv *TableView) appendItem(result gjson.Result, raw bool) {
 	newRow := table.Row{formatValue(result, raw)}
 	if len(tv.columns) > 1 && result.IsObject() {
 		newRow = make(table.Row, len(tv.columns))
+		values := result.Map()
 		for i, col := range tv.columns {
 			key := col.Title
 			if i < len(tv.columnKeys) {
 				key = tv.columnKeys[i]
 			}
-			newRow[i] = formatValue(result.Get(key), raw)
+			newRow[i] = formatValue(values[key], raw)
 		}
 	}
 	tv.table.SetRows(append(tv.table.Rows(), newRow))
@@ -639,8 +640,9 @@ func newArrayOfObjectsTableView(path string, data gjson.Result, array []gjson.Re
 
 	for _, item := range array {
 		row := make(table.Row, len(columns))
+		values := item.Map()
 		for i, key := range columnKeys {
-			row[i] = formatValue(item.Get(key), raw)
+			row[i] = formatValue(values[key], raw)
 		}
 		rows = append(rows, row)
 		rowData = append(rowData, item)
@@ -663,9 +665,10 @@ func newObjectTableView(path string, data gjson.Result, raw bool) *TableView {
 	keys := data.Get("@keys").Array()
 	rows := make([]table.Row, 0, len(keys))
 	rowData := make([]gjson.Result, 0, len(keys))
+	values := data.Map()
 
 	for _, key := range keys {
-		value := data.Get(key.Str)
+		value := values[key.Str]
 		title := SanitizeTerminalString(key.Str)
 		rows = append(rows, table.Row{title, formatValue(value, raw)})
 		rowData = append(rowData, value)
@@ -733,9 +736,10 @@ func formatValue(value gjson.Result, raw bool) string {
 func formatObject(value gjson.Result) string {
 	keys := value.Get("@keys").Array()
 	keyStrs := make([]string, len(keys))
+	values := value.Map()
 
 	for i, key := range keys {
-		val := value.Get(key.Str)
+		val := values[key.Str]
 		keyStrs[i] = formatObjectKey(SanitizeTerminalString(key.Str), val)
 	}
 
--- a/internal/jsonview/literal_keys_test.go
+++ b/internal/jsonview/literal_keys_test.go
@@ -1,0 +1,63 @@
+package jsonview
+
+import (
+	"testing"
+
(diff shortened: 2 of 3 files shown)


sylvesterkaczmarek (commented):
Checked the affected renderer/explorer paths. `@keys` still supplies object/column order, while `Map()` is now only used for exact literal lookup, so dotted or metacharacter keys can't be reinterpreted as GJSON paths. `columnKeys` is also preserved for paged rows. The static, object-table and array-of-object regressions cover the main paths. Looks consistent.

markstuart-oai:
@codex review

sylvesterkaczmarek (approved):
Re-reviewed current `8b00eeb7` after the rebase. The PR patch is materially unchanged from the head I reviewed: literal object keys are still resolved through `Map()` in the static, object-table, array-of-object, and paged-row paths, with the same regression coverage. No blocker from me.

saioai (changes requested):
i checked the updated branch. the literal a.b fix works and ten checks passed. before i approve, can you add a test for an explorer row loaded later, like the case in #121? i reproduced that path failing on main and passing with your fix locally, but the test isn’t in this PR yet. i’ll recheck after you add it.

Ultron09:
@saioai Added the regression tests covering explorer rows loaded later (both via direct append and the lazy-loaded iterator path like in #121). All tests pass locally.

sylvesterkaczmarek (approved):
Re-reviewed current `241fdd7f`. The requested lazy-load regression is now covered through `loadMoreData()` and `JSONViewer.Update()`, and the later row preserves the literal `a.b` key instead of resolving the nested `a.b` path. The direct appended-row case also pins the same behavior. This addresses the remaining test gap from the maintainer review.

Ultron09 (commented):
Hey @saioai  Please Check

saioai (approved):
the literal-key fix looks good, and the added tests cover rows loaded later, including the lazy-load path. this addresses my earlier request.

sylvesterkaczmarek:
Thanks for carrying this through and adding the later-row/lazy-load regression coverage. Glad to see #81 closed cleanly.

PR #147, opened 2026-09-01
feat(api): update usage APIs and documentation

apcha-oai:
## Summary

Updates organization usage filtering, Responses and Chat Completions usage types, and API documentation.

## Changes

- Add optional `line_items` filtering to organization cost usage requests, allowing results to be limited to exact cost line-item names
- Remove the unlaunched `compute_units` field from Responses and Chat Completions usage types
- Refresh model examples, documentation links, and reasoning-model guidance across affected API documentation


jbeckwith-oai (changes requested):
Blocking scope issue: the stated outcome is limited to removing the unlaunched `compute_units` usage fields, but this exact diff also promotes unrelated schema and public-surface changes—for example, a new organization-cost `line_items` filter, a new `max_messages` incomplete-reason value, and broad model/example/documentation updates. Those changes are not narrowly necessary for the field removal and make the patch carry behavior/API surface that the PR does not describe. Please regenerate from a baseline/config containing only the `compute_units` removal, or split and explicitly document/approve the broader promotion. I understand the compatibility/breaking-change checks triggered by the intentional field removal; that expected break is not this finding.

jbeckwith-oai (changes requested):
Thanks for broadening the description. One public-surface change is still missing from the stated scope: this diff adds `max_messages` to the Responses incomplete-reason schema in addition to the usage and documentation changes. Please explicitly include that API change in the description (or remove it), and address the two unresolved `line_items` help/example threads before approval.

pkg/cmd/adminorganizationusage.go
```diff
@@ -290,6 +290,11 @@ var adminOrganizationUsageCosts = cli.Command{
 			Default:   7,
 			QueryPath: "limit",
 		},
+		&requestflag.Flag[[]string]{
+			Name:      "line-item",
+			Usage:     "Return only costs for these exact line item names. Each value must match the complete `line_item` value, for example `gpt-5.6-sol, input_tokens`.",
+			QueryPath: "line_items",
```
HAYDEN-OAI, line 296: [P2] Preserve line_items array elements from stdin

In the existing JSON/YAML stdin mode, with no explicit `--line-item`, input such as `{"start_time":1,"line_items":["gpt-5.6-sol, input_tokens"]}` sends the JSON text of the whole array as one `line_items[]` value. The stdin adapter JSON-marshals the array before calling `Flag[[]string].Set`, whose slice setter parses and appends one string. The resulting value includes the brackets and quotes instead of the exact line-item name, so the new filter requests a different name than the caller supplied. Please preserve individual array elements when populating this query flag from stdin. Explicit quoted or repeated `--line-item` arguments are unaffected.

HAYDEN-OAI (commented):
Reviewed `0790d3f1e25b` — one concrete issue remains in the new `line_items` filter: supported JSON/YAML stdin arrays are sent as one literal JSON string rather than individual exact names. The inline comment describes the path and fix direction; explicit quoted/repeated flags are unaffected.

Static review of all 13 changed files with an independent pass; no runtime API request was sent. This comment does not resolve the separate description/scope discussion.

pkg/cmd/adminorganizationusage.go
apcha-oai: valid but existing bug; affects all arrays similarly

PR #152, opened 2026-09-03
chore: migrate to forked steady

apcha-oai:
migrate to forked steady


jbeckwith-oai (dismissed):
Three in-scope issues need addressing before approval:\n\n1. The atomic publisher is not safe under concurrent Windows/Git Bash bootstraps.  only treats  and  as contention, but Node's directory rename can report  or  when another process wins. Handle those codes as contention only when the destination now exists, then let  authenticate the winner; add focused coverage.\n2.  and the security model now claim spec inputs are local-only, while  accepts arbitrary strings and  still advertises a path or URL. Either enforce local-only input or retain the URL trust boundary in the documentation.\n3. The Deno download follows redirects without restricting redirect protocols. Please constrain the transfer and redirects to HTTPS (for example with curl protocol flags) before relying on the pinned checksum.

jbeckwith-oai (changes requested):
Three in-scope issues need addressing before approval. First, the atomic publisher is not safe under concurrent Windows and Git Bash bootstraps. The publish_directory helper only treats EEXIST and ENOTEMPTY as contention, but Node directory rename can report EPERM or EACCES when another process wins. Handle those codes as contention only when the destination now exists, then let steady_verify authenticate the winner, and add focused coverage. Second, CONTRIBUTING.md and the security model now claim spec inputs are local-only, while scripts/mock accepts arbitrary strings and scripts/test still advertises a path or URL. Either enforce local-only input or retain the URL trust boundary in the documentation. Third, the Deno download follows redirects without restricting redirect protocols. Please constrain the transfer and redirects to HTTPS before relying on the pinned checksum.

jbeckwith-oai (approved):
Re-reviewed the requested fixes. Concurrent Windows publication is guarded and covered, Deno downloads and redirects are restricted to HTTPS, and the documentation and security model now accurately retain trusted URL input as an explicit untrusted-content boundary. Exact-head checks are green.

PR #153, opened 2026-09-03
chore(api): document incomplete web search call status

markstuart-oai:
## Summary

The API reference lists incomplete as a web-search call status.

## Changes

- Add incomplete to the web-search status reference.


HAYDEN-OAI (approved):
Reviewed the complete generated-artifact diff and the relevant stable/beta web-search schemas and CLI consumers. Both status references resolve to the same five-value contract, including `incomplete`; the transformed-spec hash matches the file, generated command output is unchanged, and the CLI response paths preserve the status. YAML parsing/reference checks and `git diff --check` are clean.

PR #157, opened 2026-09-05
feat(api): add service-account API key expiration fields

jbeckwith-oai:
## Summary

Adds optional expiration fields for service-account API keys and clarifies the Responses store option.

## Changes

- Allow service-account API key creation requests to specify expires_in_seconds.
- Expose expires_at on service-account API key responses.
- Clarify the default and retention behavior of the Responses store option.


markstuart-oai (approved):
Reviewed all six changed files and the shared request-building path. The expiration flag uses the existing nullable integer abstraction, preserves omission versus explicit null, and matches the pinned OpenAPI schema. The new field remains in the generated command boundary without adding special-case orchestration. No substantive issues found.

Validation: focused requestflag/apiquery tests, go vet ./..., CLI build, gofmt, and diff whitespace checks passed. A local HTTP harness verified omitted/numeric/null expiration values, YAML input and flag precedence, invalid-integer rejection before HTTP, and preservation of numeric/null expires_at responses. The full mock-server suite was not run.

PR #160, opened 2026-09-06
fix(jsonview): retain lazy-loaded items across JSON toggles

markstuart-oai:
## Summary

After the JSON explorer loads additional items, pressing `r` rebuilds the view from its original preload and discards the loaded rows while keeping the advanced iterator. Keep array views and render toggles from all accumulated rows so repeated toggles and continued loading retain item order without skips or duplicates.

Iterator commands now return results to the UI update loop. This keeps an in-flight load attached to its original view across toggles and nested navigation, and formats its result using the current display mode. Existing nonstreaming rendering and terminal escaping remain unchanged; no dependencies or generated sources change.

## Validation

- The new key-message retention regression fails on unchanged main and passes with the fix.
- Focused race tests cover repeated toggles, continued loading/exhaustion, blocked in-flight loads, nested navigation/back, terminal escaping, and nonstreaming controls.
- `go test ./internal/...`
- `go test ./... -run '^$'`
- `go mod verify`
- `./scripts/lint` (build)
- Windows amd64 test cross-compilation; no Windows runtime or full interactive-terminal claim.
- Full `./scripts/test` passed against the pinned Steady local mock after direct loopback health verification, including the script's Windows cross-compilation phase. Synthetic credentials only.

### DIFF
--- a/internal/jsonview/explorer.go
+++ b/internal/jsonview/explorer.go
@@ -138,65 +138,56 @@ func (tv *TableView) Update(msg tea.Msg, raw bool) tea.Cmd {
 		// Load more when we're at the last row
 		if cursor == totalRows-1 {
 			tv.isLoading = true
-			return tv.loadMoreData(raw)
+			return tv.loadMoreData()
 		}
 	}
 
 	return cmd
 }
 
-func (tv *TableView) loadMoreData(raw bool) tea.Cmd {
+// tableItemMsg transfers iterator work back to the UI loop. Commands must not
+// mutate a view while key and resize messages are being processed.
+type tableItemMsg struct {
+	view   *TableView
+	result gjson.Result
+	err    error
+}
+
+func (tv *TableView) loadMoreData() tea.Cmd {
+	iterator := tv.iterator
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the exact head and found no blocking issues. Moving iterator work into a `tableItemMsg` removes command-goroutine mutation of `TableView`, while retaining the originating view pointer lets the UI loop safely complete a pending load after raw-mode toggles, resizes, or nested navigation. Array toggles now rebuild only presentation from accumulated `rowData`, preserving the iterator, loaded rows, order, and view identity; the result is formatted using the display mode current when it is applied. Exhaustion and marshal failures clear `isLoading`, and stale results are ignored if their view is no longer in the stack. Focused tests cover repeated toggles, continued loading/exhaustion without skips or duplicates, blocked in-flight loads, nested navigation/back, resize, terminal-control escaping, and nonstreaming rendering. All CI, lint, tests, CodeQL, Castiron checks, and exact-head code/security reviews are green.

PR #162, opened 2026-09-07
fix(requestflag): handle inner fields after null array objects

markstuart-oai:
## Problem and fix

An inner field flag after a null array-object value panics while assigning into a nil map. Initialize that trailing map and store it in the existing slice slot before assigning the field.

With a fresh CLI build and an empty environment, this parser-only command previously exited 2 with `assignment to entry in nil map`; it now exits 0 and produces identical help to the `{}` control:

```sh
env -i PATH=/usr/bin:/bin /tmp/openai chat:completions create --function null --function.name demo --help
```

This is a parsing fix, not a claim that null alone is a valid API function schema. Explicit null without a later inner assignment remains null. Earlier elements, sibling merging, repeated-field creation of a new element, typed and seeded untyped arrays, and per-trailing-element CLI/stdin precedence remain intact. Generated command sources are unchanged.

## Regression coverage and validation

Added command-parsing/request-body regressions plus trailing-element stdin precedence and seeded untyped-array coverage. The new CLI regression failed at the nil-map assignment before the fix. Two consecutive independent review rounds were clean on the unchanged diff.

Passed:
- `go test ./internal/requestflag -count=1`
- `go test ./internal/...`
- `go test ./... -run '^$'` (compile-only)
- `go mod verify`
- `./scripts/lint` (Go build, not comprehensive static analysis)
- Fresh CLI build and the parser-only reproduction/control above
- `GOOS=windows GOARCH=amd64 go test -c -o /tmp/requestflag.test.exe ./internal/requestflag` (compilation only)
- `git diff --check`

Tests used synthetic local inputs and made no live API calls. Full mock integration and Windows runtime execution were not run. The committed-candidate custom-code budget check passed: +427/-68 = 495 lines against a 1000-line limit, with verified generated baselines.

### DIFF
--- a/internal/requestflag/innerflag_test.go
+++ b/internal/requestflag/innerflag_test.go
@@ -1,9 +1,12 @@
 package requestflag
 
 import (
+	"context"
+	"encoding/json"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"github.com/urfave/cli/v3"
 )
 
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Reviewed exact head e339ba9e. The fix initializes only a nil trailing map and writes it back into the existing slice slot before assigning the inner field, which removes the panic while preserving explicit null alone, sibling merging, repeated-field element creation, earlier elements, and stdin precedence. The added typed/untyped regressions cover those boundaries. Focused go test, gofmt, go vet, and diff checks passed locally; the full hosted CI/CodeQL/Castiron set is green with no unresolved threads. Focused request-parsing security review found no new data exposure, unsafe coercion, or cross-element mutation.

PR #163, opened 2026-09-07
fix(jsonview): preserve raw API fields in lazy-loaded rows

markstuart-oai:
## Problem and fix

Lazily loaded explorer rows re-marshal SDK response values, losing unknown response fields and turning an absent or null `shutdown_date` into `"0001-01-01T00:00:00Z"`. Preloaded rows already preserve those fields using `RawJSON()`.

Prefer the existing `RawJSON()` interface for each lazy-loaded item, matching preload and ordinary iterator output. Ordinary `json.Marshal` fallback and marshal-error delivery remain unchanged. Commands still return `tableItemMsg`; the UI update loop alone appends rows, preserving retention, navigation, iterator order, and nested/in-flight delivery across raw/formatted toggles.

## Regression and validation

The new regression unmarshals synthetic response JSON into the pinned `openai-go/v3` `Model` and exercises the current command-to-UI message flow. Both absent and null cases failed on main before the fix, showing the lost unknown field and synthesized timestamp; both now pass. It also covers repeated toggles, selected content/navigation, ordinary values, and marshal errors. Existing retention, in-flight/nested delivery, navigation, and terminal-sanitization tests pass. Two consecutive independent review rounds were clean on the unchanged diff.

Passed:
- `go test ./internal/jsonview -count=1`
- `go test ./internal/...`
- `go test ./... -run '^$'` (compile-only)
- `go mod verify`
- `./scripts/lint` (Go build)
- `go test -race ./internal/jsonview -count=1`
- `GOOS=windows GOARCH=amd64 go test -c -o /tmp/jsonview.test.exe ./internal/jsonview` (compilation only)
- `git diff --check`

The full local `./scripts/test` workflow was deliberately not run: its mock commands capture non-TTY output and do not exercise interactive lazy loading. Direct SDK/UI component regressions and race tests cover the changed boundary. No local mock integration, PTY, live API, or Windows runtime testing is claimed.

The committed-candidate custom-code budget check passed: +427/-68 = 495 lines against a 1000-line limit, with five mixed files and verified generated baselines. Only handwritten explorer source and its existing regression suite changed.

### DIFF
--- a/internal/jsonview/explorer.go
+++ b/internal/jsonview/explorer.go
@@ -164,7 +164,12 @@ func (tv *TableView) loadMoreData() tea.Cmd {
 			msg.err = iterator.Err()
 			return msg
 		}
-		jsonBytes, err := json.Marshal(iterator.Current())
+		item := iterator.Current()
+		if hasRaw, ok := item.(hasRawJSON); ok {
+			msg.result = gjson.Parse(hasRaw.RawJSON())
+			return msg
+		}
+		jsonBytes, err := json.Marshal(item)
 		msg.err = err
 		if err == nil {
 			msg.result = gjson.ParseBytes(jsonBytes)
--- a/internal/jsonview/explorer_stream_test.go
+++ b/internal/jsonview/explorer_stream_test.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/charmbracelet/bubbles/help"
 	tea "github.com/charmbracelet/bubbletea"
+	"github.com/openai/openai-go/v3"
 	"github.com/stretchr/testify/require"
 	"github.com/tidwall/gjson"
 )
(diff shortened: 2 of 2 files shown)


jbeckwith-oai (approved):
Reviewed exact head 10f0eda3. Reusing the existing RawJSON contract in the lazy-load command makes streamed rows match preload/iterator output, preserving unknown fields and absent/null semantics without moving UI mutation out of the Bubble Tea update loop. Ordinary json.Marshal fallback, marshal-error delivery, iterator ordering, nested/in-flight completion, selection, navigation, and raw/formatted toggles remain covered. Focused go test (including -race), gofmt, go vet, and diff checks passed locally; hosted CI/CodeQL/Castiron is green with no unresolved threads. Focused security review confirmed the path retains existing gjson parsing and terminal sanitization boundaries.

PR #164, opened 2026-09-07
fix(autocomplete): use distinct status for no completion

markstuart-oai:
## Problem and result

Completing a non-file flag value such as `--format` or `models list --max-items` suggested unrelated filenames. The omitted Go constant expression made `ShellCompletionBehaviorNoComplete` repeat file status `10`, although all four bundled shell scripts already treat `11` as no completion.

Assign `NoComplete` explicitly to `11`, preserving Default `0` and File `10`. The Bash helper invocation also copies completion words individually into a quoted array: Bash 3.2 collapsed the previous array slice into one argument, preventing nested commands from reaching the intended no-completion behavior. Spaces and empty arguments are preserved. Inherited global-flag lookup/enumeration, colon reconstruction and hidden-flag behavior remain outside this change.

## Regression and validation

The subprocess regression exercises the real completion entrypoint and literal wire statuses, rendered Bash candidates with synthetic filename fixtures, and exact NUL-delimited helper arguments. File, explicit `@`, default-status, spaced-value and empty-value controls are covered. Two consecutive independent review rounds were clean on the unchanged final diff.

Passed on the final source:
- Full autocomplete suite and `go test -race ./internal/autocomplete/... -count=1` on Linux.
- Native macOS Bash 3.2.57 / Go 1.27.0 full autocomplete race suite, including all seven protocol cases. The original nested failure was reproduced before the repair.
- `go test ./internal/... -count=1`.
- `go test ./... -run '^$' -count=1` (compile only).
- `go mod verify`.
- `./scripts/lint` (Go build only).
- `go vet ./internal/autocomplete/...`, formatting, diff and Bash syntax checks.
- Windows amd64 test cross-compilation on Linux and separately on macOS; Windows runtime tests were not run.
- Fresh real CLI Bash and PowerShell synthetic probes preserving non-file suppression and file/explicit-`@` behavior.
- Trusted-main custom-code report/budget against the actual follow-up commit: verified generated snapshot, five unchanged mixed files, 495/1000 lines.

The native macOS full `./scripts/test -count=1` started the approved pinned Steady mock and executed the full suite, but **failed** solely at `TestFilesCreateCLICancelClosesStalledFIFO` (`multipartresource_unix_test.go:129`, expected broken pipe, got nil). The same isolated test failed identically on the immutable pre-repair commit and repaired source in the same environment. Its precise cause is undetermined; this PR does not alter or skip it. All other packages passed. Windows compilation was run separately because the full script stopped at that failure.

The Linux local full-mock run remains environmentally unverified because the pinned JSR dependency is unavailable in its verified cache and the normal bootstrap fetch failed. [CI run 34166447608](https://github.com/openai/openai-cli/actions/runs/34166447608) passed on head `16e05075f292b77658df7523baf65e5a99b6352e`, including the full Linux Steady mock suite, Windows test cross-compilation, and all nine artifact build targets.

Bash 3.2 rendered tests use the existing test-only `mapfile` shim. They establish argument-boundary/protocol behavior, not universal unshimmed stock Bash 3.2 completion support. Fish and Zsh were source-reviewed; native Fish execution was unavailable. No live API or customer data was used.

Compatibility: this restores the bundled scripts' existing protocol. Third-party completion consumers relying on the erroneous status `10` for non-file values will now receive `11`.

### DIFF
--- a/internal/autocomplete/autocomplete.go
+++ b/internal/autocomplete/autocomplete.go
@@ -87,9 +87,9 @@ func NewShellCompletion(name string, usage string) ShellCompletion {
 type ShellCompletionBehavior int
 
 const (
-	ShellCompletionBehaviorDefault ShellCompletionBehavior = iota
-	ShellCompletionBehaviorFile                            = 10
-	ShellCompletionBehaviorNoComplete
+	ShellCompletionBehaviorDefault    ShellCompletionBehavior = iota
+	ShellCompletionBehaviorFile                               = 10
+	ShellCompletionBehaviorNoComplete                         = 11
 )
 
 type CompletionResult struct {
--- a/internal/autocomplete/protocol_test.go
+++ b/internal/autocomplete/protocol_test.go
@@ -8,6 +8,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"slices"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -63,6 +64,8 @@ func TestShellCompletionProtocol(t *testing.T) {
 		{"root value", []string{"--format", "candidate-"}, 11, "", ""},
 		{"nested local value", []string{"models", "list", "--max-items", "candidate-"}, 11, "", ""},
 		{"file value", []string{"--file", "candidate-"}, 10, "", "candidate-fixture.txt\n"},
+		{"spaced preceding value", []string{"--format", "two words", "--file", "candidate-"}, 10, "", "candidate-fixture.txt\n"},
+		{"empty preceding value", []string{"--format", "", "--file", "candidate-"}, 10, "", "candidate-fixture.txt\n"},
 		{"explicit file prefix", []string{"--format", "@candidate-"}, 11, "", "@candidate-fixture.txt\n"},
 		{"command prefix", []string{"mo"}, 0, "models\n", "models\n"},
 	} {
@@ -103,6 +106,7 @@ if ! type mapfile >/dev/null 2>&1; then
   }
 fi
 openai() {
+  printf '%s\0' "$@" > helper.argv
   "$test_binary" -test.run='^TestShellCompletionProtocolHelper$' -- openai "$@"
 }
 ` + script + `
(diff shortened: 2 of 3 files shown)


markstuart-oai:
@codex review

markstuart-oai:
@codex security review

jbeckwith-oai (changes requested):
[P2] Keep nested completion arguments separate on Bash 3.2

The new protocol test fails deterministically on the Bash version bundled with macOS (GNU Bash 3.2.57): TestShellCompletionProtocol/nested_local_value/rendered_bash gets "\n" instead of no candidates. The direct helper invocation exits 11 correctly, but bash_autocomplete.bash expands the preceding COMP_WORDS slice as one argument ("models list --max-items") under Bash 3.2 while local IFS is newline. The helper therefore returns status 0, and the script emits one empty completion. This is within the supported path because the test itself installs a Bash 3 mapfile fallback.

Please preserve array-element boundaries for the helper invocation and keep this regression running on Bash 3.2 (or otherwise exercise that supported behavior). The root-value case passes; the nested local-flag case remains broken.

Exact head: 3525e08a9a0e0b052d310ec481107667348a164b. gofmt and go vet pass. go test -race ./internal/autocomplete/... fails only on the new nested rendered-Bash assertion above.

markstuart-oai:
Addressed the Bash 3.2 argument-boundary finding from [your review](https://github.com/openai/openai-cli/pull/164#pullrequestreview-5135316198) in 16e05075f292b77658df7523baf65e5a99b6352e. The helper now receives individually copied, quoted array elements; the regression checks exact argv, including spaced and empty values.

The original nested failure reproduced on native Bash 3.2.57, and the repaired full autocomplete race suite passed there with all seven protocol cases (3.838s). Existing file/`@` controls pass; rendered tests retain the existing `mapfile` shim.

The native full mock suite did run but failed only the FIFO cancellation assertion at `multipartresource_unix_test.go:129`; the immutable pre-repair commit and repaired source fail identically in that environment. Its cause remains undetermined, and this change does not touch or skip it. New-head full Linux mock CI and bot reviews are pending.

jbeckwith-oai (approved):
Re-reviewed exact head 16e05075f292b77658df7523baf65e5a99b6352e. The Bash 3.2 argument-slicing blocker is fixed by building the completion argv one indexed word at a time, preserving spaces and empty values. The protocol regression now verifies the exact NUL-delimited argv received by the helper. On macOS Bash 3.2.57, the focused protocol test and full autocomplete package passed; gofmt, go vet, and diff checks were clean. No remaining in-scope findings.

PR #166, opened 2026-09-08
fix(autocomplete): complete inherited flags after subcommands

markstuart-oai:
## Summary

Flag completion stopped at the current command's declarations, so `openai __complete -- models list --fo` returned no suggestions even though `--format` and `--format-error` are valid there. Include inherited flag names and aliases after nested commands, and consume their values before continuing command traversal.

Use the pinned urfave inheritance contract: preserve child name/alias overrides, nearest-ancestor precedence, hidden inherited flags, and flags marked local. Preserve completion statuses, file completion and explicit `@` paths. Completion does not execute API requests. Regression coverage exercises raw and initialized command trees, the production entrypoint, and rendered Bash completion.

## Validation

Passed locally with Go 1.27.0 on Linux (module version and dependencies unchanged):

- `go test ./internal/autocomplete ./cmd/openai -count=1`, `go test ./internal/...`, and `go test ./... -run '^$'`.
- `go mod verify`, `./scripts/lint` (build only), `go test -race ./internal/autocomplete ./cmd/openai -count=1`, and `go vet ./internal/autocomplete ./cmd/openai`.
- `GOOS=windows GOARCH=amd64 go test -c -o /tmp/openai-cli-windows-tests/ ./...` (output directory abbreviated here); no Windows runtime test.
- Changed-file formatting and diff checks; rendered Bash execution; Bash, Zsh and PowerShell syntax checks. Fish was unavailable; Zsh and PowerShell runtime behavior was not tested.
- Fresh before/after binaries: nested `--fo` changes from empty output to `--format` and `--format-error`; root output stays the same. All four probes exit 0 with empty stderr. The initial inherited-completion regression failed on the original implementation.

**Full `./scripts/test` mock suite NOT RUN locally:** the pinned Steady bootstrap failed loading the JSR `@std/cli` manifest. A direct connectivity probe and an existing local health response did not establish a working mock server. No network or security controls were bypassed. Full CI/mock-suite validation remains required.

### DIFF
--- a/cmd/openai/completion_test.go
+++ b/cmd/openai/completion_test.go
@@ -1,0 +1,67 @@
+package main
+
+import (
+	"bytes"
+	"os"
+	"os/exec"
+	"slices"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestInheritedCompletionMainHelper(t *testing.T) {
+	if os.Getenv("OPENAI_CLI_INHERITANCE_HELPER") != "1" {
+		return
+	}
+	os.Args = os.Args[slices.Index(os.Args, "--")+1:]
+	main()
+	os.Exit(0)
+}
+
+func TestInheritedCompletionMain(t *testing.T) {
+	binary, err := os.Executable()
+	require.NoError(t, err)
+	// Keep real credentials and environment-selected client configuration out of
+	// subprocesses; completion never needs an API request or credential values.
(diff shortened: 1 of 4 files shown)


saioai (changes requested):
inherited flag completion is already on main, but the fix for values that look like flags is still useful. please rebase this onto main and narrow it to that fix and its tests, preserving the completion behavior already there.

PR #172, opened 2026-09-08
fix(output): apply transforms to interactive list items

markstuart-oai:
## Problem and result

Interactive list output ignored `--transform`: for example, `files list --format explore --transform id` displayed and printed whole file objects. Apply the existing GJSON projection once to each original item before it becomes explorer row data, for both initially loaded and subsequently fetched items. Printing, navigation, and raw-mode toggles now use the projected value.

Preserve missing-path fallback, explicit null and other scalar/object/array values, SDK `RawJSON()` fields, serialization errors, and lazy loading across pending UI transitions. Empty-object rows use the existing generic Items presentation when there are no object columns, keeping later values visible. Existing formats, raw response handling, dependencies, generated commands, and pagination limits are unchanged.

## Validation

- Reproduced on clean main and verified the fix with the normal CLI against a synthetic localhost two-page server in a real PTY, including a genuinely lazy second-page row, printing, toggles, and nested navigation. Separate production-output helper instrumentation verified iterator call counts without eager consumption.
- Regression tests cover once-only projection, missing versus explicit values, unknown SDK fields, marshal/upstream errors, initial and lazy rows, pending loads, terminal sanitization, and empty-object rows followed by other values.
- Passed focused tests and race tests, `go test ./internal/...`, `go test ./... -run '^$'`, `go mod verify`, `go vet ./pkg/cmd ./internal/jsonview`, and `./scripts/lint` (build only). Terminal controls and 22 baseline/candidate nonterminal comparisons passed.
- All-package Windows test cross-compilation passed; native Windows execution was not performed. Local runtime was Go 1.27.0; the module minimum remains Go 1.25.0.
- Trusted-main custom-code budget check passed against the actual committed candidate: 495/1000 lines, with no budget or generation-policy changes.

The full local mock suite did not run: the supported Steady launcher returned `Missing or linked Steady cache`. The denied setup route was not retried. Full hosted CI/mock validation remains required. All runtime probes used synthetic inputs; no live API calls, real credentials, or customer data were used. SDK CODEOWNER review remains required.

### DIFF
--- a/internal/jsonview/explorer.go
+++ b/internal/jsonview/explorer.go
@@ -117,6 +117,7 @@ type TableView struct {
 	table      table.Model
 	rowData    []gjson.Result
 	iterator   AnyIterator
+	transform  string
 	isLoading  bool
 	columns    []table.Column
 	columnKeys []string
@@ -155,6 +156,7 @@ type tableItemMsg struct {
 
 func (tv *TableView) loadMoreData() tea.Cmd {
 	iterator := tv.iterator
+	transform := tv.transform
 	return func() tea.Msg {
 		msg := tableItemMsg{view: tv}
 		if iterator == nil {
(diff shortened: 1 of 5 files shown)


saioai (changes requested):
i checked this at 2a031a2. the transform works for initial and lazy rows, but lazy rows still keep the full original JSON in memory through result.Raw even when the selected value is tiny. please clone the projected raw string, cover the retention case, and update the branch from main. i'll recheck the new commit.

PR #177, opened 2026-09-09
feat(api): Add service account key expiration flag

apcha-oai:
## Summary

Allow administrators to set the initial API key lifetime when creating a project service account. Refresh CLI help links to point to current developer documentation.

## Changes

- Add the optional `--expires-in-seconds` flag to project service account creation, including policy-limit and service-account-only usage guidance.
- Update command and flag help links across API resources to use current developer documentation and API reference pages.


pkg/cmd/adminorganizationprojectserviceaccount_test.go
```diff
@@ -18,14 +18,16 @@ func TestAdminOrganizationProjectsServiceAccountsCreate(t *testing.T) {
 			"--project-id", "project_id",
 			"--name", "name",
 			"--create-service-account-only=true",
+			"--expires-in-seconds", "1",
```
markstuart-oai, line 21: [P3] Exercise expiration with a service account that creates a key

Both new cases combine a non-null expiration with `create_service_account_only: true`, even though the documented contract rejects that combination because no key is created. The mock helper only asserts command success, so these fixtures do not demonstrate the supported expiration workflow. Please omit `create_service_account_only` (or set it to false) in both the flag and piped-YAML cases, and keep keyless creation as a separate case with expiration omitted/null. Fix the generated fixture source so regeneration preserves the valid combinations.

markstuart-oai (approved):
Reviewed the expiration contract, request serialization, generated schema/help changes, and surrounding command structure. The new flag reuses the existing nullable-integer path without adding special-case logic. Local synthetic requests preserve omitted/null/value inputs and expiration response fields; all packages compile and request-flag tests pass. One low-priority fixture correction is noted inline.

jbeckwith-oai (dismissed):
Two in-scope blockers remain on this exact head. The new CLI test combines --expires-in-seconds with --create-service-account-only=true even though the command contract forbids that combination, so it does not cover a successful expiring-key request. Also, the “available models” link in pkg/cmd/moderation.go points back to the moderation guide rather than the models documentation. Both are already called out in the existing inline review threads; please correct the source and regenerate.

PR #180, opened 2026-09-09
ci: resolve fork PRs missing Castiron run associations

marcuswood-oai:
## Summary

When GitHub omits a fork PR from both the workflow-run association and the base repository's commit-to-PR endpoint, Castiron cannot resolve the PR and leaves the required budget statuses missing.

Port the focused fix from https://github.com/openai/openai-python/pull/3831: fall back to listing open PRs by the source run's fork owner and branch, then retain the existing exact-head, target, freshness, and ambiguity checks. Cover report computation, budget evaluation, and publication; update the reporter integrity digest.

## Validation

- Both evaluator and status-publisher regressions reproduce on unpatched main.
- Castiron reporter/budget suite: 53 tests, OK (1 optional compiler-contract test skipped), including JavaScript publishers, pagination, stale heads, wrong targets, and ambiguous matches.
- Ruff and Pyright for `scripts/castiron`; `git diff --check`.
- Patched lookup verified against https://github.com/openai/openai-cli/pull/43 with read-only GitHub requests.

The trusted handler runs from main. After merging, rerun affected Castiron custom-code workflows so the required statuses can be published.

### DIFF
--- a/.github/workflows/castiron-custom-code-comment.yml
+++ b/.github/workflows/castiron-custom-code-comment.yml
@@ -145,8 +145,11 @@ jobs:
             const {data: main} = await github.rest.git.getRef({...context.repo, ref: 'heads/main'});
             const base = main.object.sha;
             if (run.event === 'pull_request') {
-              const pulls = run.pull_requests.length ? run.pull_requests : await github.paginate(
+              let pulls = run.pull_requests.length ? run.pull_requests : await github.paginate(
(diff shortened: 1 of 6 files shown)


saioai (approved):
This makes sense. It fixes the missing PR link we saw on #181 and still checks the exact commit before posting results. Looks good to me.

PR #181, opened 2026-09-10
docs: correct bootstrap dependency check description

Tmwakalasya:
`CONTRIBUTING.md` describes bootstrap as running `go mod tidy` and potentially rewriting dependency files. Update it to document `go mod tidy -diff`: when the manifests need changes, the check prints a diff and exits without rewriting them. Dependency download and verification follow a successful check.

Validation:

- `go test ./scripts -run '^TestBootstrap' -count=1` (passed)
- `git diff --check` (passed)


saioai (approved):
Description matches scripts/bootstrap: it checks module files without rewriting them, then downloads and verifies dependencies. Looks good to me.

PR #186, opened 2026-09-14
test: isolate flag instances in parallel TestFlagSet subtests

Tmwakalasya:
## Summary

The parallel subtests in `TestFlagSet` shared string and integer `Flag` instances. Both `PreParse` and `Set` mutate flag state, so concurrent subtests raced on the same objects.

- Construct fresh string and boolean flags inside their subtests.
- Use a small factory to give each integer subtest its own flag while preserving the existing validator.
- Keep all existing assertions and `t.Parallel()` calls. No production code or dependency files change.

## Validation

Before the fix, this command failed with `WARNING: DATA RACE`:

```sh
go test -race ./internal/requestflag -run '^TestFlagSet$' -count=1
```

After the fix, all of these checks passed:

```sh
go test -race ./internal/requestflag -run '^TestFlagSet$' -count=100 -shuffle=on
go test -race ./internal/requestflag -count=1
go vet ./internal/requestflag
go mod verify
git diff --check
```

The full integration/mock-server suite was not run for this test-only change.


saioai (approved):
looks good to me. the race showed up on main, and this passed 100 runs with the race detector on this branch. checks are green too.

PR #190, opened 2026-09-15
feat(cli): support custom request headers

markstuart-oai:
## Summary

Add repeatable `--header` / `-H 'Name: Value'` support to API commands. Values stay literal, duplicate names use the last value regardless of case, and explicit headers override default and endpoint headers.

Document newline-separated `OPENAI_CUSTOM_HEADERS` for supplying headers through the environment. Include headers from both inputs in request/response debug redaction, keep values out of help, and support inherited flags in shell completion. The change uses handwritten CLI code and preserves generated sources and dependencies.

## Validation

- CLI and internal tests, including local HTTP coverage for GET, JSON, streaming, uploads, API/admin authentication, header precedence, malformed input, help, and completion.
- Environment-only API/admin requests, explicit header overrides, and request/response redaction; the environment-redaction regression was reproduced before the fix.
- Cross-origin regression verifies the configured origin receives the header and the redirect target receives no request.
- All-package compilation, `scripts/lint`, `go vet ./...`, `go mod verify`, and Windows test compilation.
- Two independent adversarial reviewers per round, with two final clean rounds.


jbeckwith-oai (approved):
Reviewed the exact head, including header parsing and precedence, debug-log redaction, redirect behavior, inherited completion semantics, and focused regression coverage. Local focused tests, race tests, compilation, lint, vet, and module verification passed; CI and the exact-head security review are green.

PR #195, opened 2026-09-16
fix(requestflag): preserve JSON numbers written in exponent form

thegoodengineer:
## Summary

JSON numbers written with an exponent and no fraction, such as `1e-05`, are sent to the API as strings when they come from piped stdin or a JSON flag value. Python, JavaScript, and Go all write small floats this way, so ordinary generated JSON is affected.

## Problem

Both YAML/JSON decode sites, `flagOptions` for piped stdin and the default branch of `parseCLIArg` for flag values, use go-yaml. go-yaml only classifies a scalar as a float when it contains a decimal point (`toNumber` in `token/token.go` checks `dotCount == 1`), so `1e-05` fails integer parsing and falls back to a string. This is open upstream as goccy/go-yaml#916 and is still present in v1.19.2 and on master.

Request bodies captured with a local echo server, `main` compared with this branch:

| Input | `main` sends | This branch sends |
| --- | --- | --- |
| stdin from Python `json.dumps({"top_p": 0.00001})`, which writes `1e-05` | `"top_p":"1e-05"` | `"top_p":0.00001` |
| stdin from Node `JSON.stringify({temperature: 1e-7})`, which writes `1e-7` | `"temperature":"1e-7"` | `"temperature":1e-7` |
| `--text '{"format":{"type":"json_schema","name":"score","schema":{"type":"number","minimum":1e-7}}}'` | `"minimum":"1e-7"` | `"minimum":1e-7` |
| stdin to `vector-stores:files update` from Python `{"attributes":{"min_relevance":0.00005}}` | `"min_relevance":"5e-05"` | `"min_relevance":0.00005` |

For typed parameters such as `top_p`, the API receives a string where it expects a number, even though the input clearly contains one. Free-form values have no type to check against, so the change goes through silently: the structured output schema above carries a string `minimum`, and the vector store attribute is sent as a string instead of a number.

The same thing happens to piped maps assigned to query, header, or path flags. `formatForFlagSet` re-encodes them with `encoding/json`, which writes `float64(1e-07)` as `1e-7`, and the flag parser turns that back into a string.

Numbers that already have a fraction, like `2.5e-05`, decode correctly, which makes this easy to miss.

## Fix

Add `requestflag.UnmarshalYAMLOrJSON` and use it at both decode sites. When the input is valid JSON, it inserts a zero fraction into exponent numbers that lack one (`1e-05` becomes `1.0e-05`), which go-yaml decodes to the identical `float64`. The scan skips string contents, including escaped quotes and backslashes, so keys and strings that look like numbers are never touched.

I kept go-yaml in charge of decoding rather than switching valid JSON to `encoding/json`, because that would change behavior this fix has no reason to touch: the `uint64`/`int64` integer types that existing code and tests rely on, duplicate key rejection, and handling of out-of-range integers. As a result:

- YAML and invalid input reach go-yaml byte for byte. YAML is left alone on purpose, since `1e-05` is a string under YAML 1.1 and a float under YAML 1.2, which is the question in goccy/go-yaml#916. JSON has a single meaning.
- Integers keep their existing Go types, and duplicate keys are still rejected.
- Out-of-range values such as `1e400` decode the same way `1.5e400` already does.
- Valid JSON with nothing to rewrite is returned as the original slice, without a copy.
- A bare exponent on a union-typed flag, for example `--input 1e5`, is now a number. That matches `--input 42` and `--input 1.5`, which `main` already sends as numbers.

`pkg/cmd/flagoptions.go` no longer imports go-yaml directly. No dependency, generated, schema, or budget files change, and every production change is in handwritten code.

## Tests

- `internal/requestflag/unmarshal_test.go` covers each producer's exponent form, uppercase and signed exponents, negative and zero mantissas, a top-level number, nested schemas, look-alike keys and strings (including `["\\",1e2]`, a string ending in an escaped backslash followed by a number), existing integer types, pass-through of YAML and malformed input, and duplicate key rejection. It also covers all three flag paths: a map flag, a repeated object flag, and a piped map re-parsed through `ApplyStdinDataToFlags`.
- `cmd/openai/main_jsonnumbers_test.go` runs the real entrypoint with a JSON request on stdin plus a `--text` schema, with `OPENAI_UNTRUSTED_STDIN` both off and on, and checks that the captured request body contains JSON numbers. `runMainDispatchWithEnv` now delegates to a `runMainDispatchWithStdin` variant for this. The test passes a regular file as stdin so it does not depend on pipe timing in `isInputPiped`.

With the production code from `main`, the end-to-end test fails with:

```
request top_p = "1e-05" (string), want JSON number 1e-05
request text.format.schema.minimum = "1e-7" (string), want JSON number 1e-07
```

With the rewrite disabled, every exponent case in the unit tests fails the same way, for example `expected: map[string]interface {}{"top_p":1e-05}` / `actual: map[string]interface {}{"top_p":"1e-05"}`.

## Validation

Linux (WSL Ubuntu 22.04) with Go 1.25.14, on a clean clone of this commit, all passing:

- `go mod verify` and `go mod tidy -diff`
- `go vet ./...`
- `./scripts/lint`
- `go test ./... -run '^$'`
- `go test ./internal/...` and `go test ./cmd/openai`
- `go test ./pkg/cmd -run 'Stdin|EmbedFiles|IsUTF8|Multipart'`, the existing piped stdin, file embedding, and multipart tests
- `go test -race ./internal/requestflag ./cmd/openai`
- `go test -race -count=100 -shuffle=on ./internal/requestflag -run 'TestUnmarshalYAMLOrJSON|TestJSONExponentNumbersInFlagValues'`
- `go test -race -count=50 ./cmd/openai -run '^TestMainJSONExponentNumbers$'`
- `gofmt -l` on the changed files

I did not run the full `./scripts/test` suite against the Steady mock server; that is left to CI. On Windows with Go 1.27.1 the new tests also pass. `internal/autocomplete` fails there on an unmodified `main` too, because it shells out to bash, and it passes on Linux.

I also fuzzed the rewrite locally for about 1.4 million inputs (the fuzz target is not part of this PR). The checks were: output is valid JSON whenever the input is, it decodes to the same values with `encoding/json` using `UseNumber`, the only change is `.0` inserted before an exponent, and non-JSON input comes back unchanged. There were no failures.

On a 6.8 MB piped JSON body, go-yaml decoding takes about 500 ms. The added pass takes about 9 ms when there is nothing to rewrite and about 30 ms with 60,000 numbers rewritten.


saioai (approved):
the json number bug is reproduced on main and fixed here. the request and compatibility checks pass.

PR #196, opened 2026-09-17
feat(api): add prompt-cache prewarming

jbeckwith-oai:
## Summary

Adds the optional prompt-cache prewarming flag to stable and beta response creation commands.

## Changes

- Support `--prompt-cache-options.prewarm` and the equivalent object and YAML options
- Preserve omitted options and explicitly supplied `false` and `true` values


markstuart-oai (approved):
Reviewed `8b69add54b04dd6bc07f6841eb57cf55d141931f`, including an independent CLI review and the complete supporting schema refresh. No actionable findings. Stable and beta creation both register the optional boolean through the existing inner-flag machinery; the schema definitions agree with the canonical optional field. The source trace preserves omission, explicit false/true, sibling cache options and caller-supplied request data. Server-side prewarm behavior is not reimplemented in the CLI, and no new abstraction or shared parsing change is introduced.

[CI run 35262485829](https://github.com/openai/openai-cli/actions/runs/35262485829) passed build/artifact checks, lint, the CLI/internal suite against Steady, and Windows checks. Its tested merge `68fef721b8c67c5dacf98ead39cc17338ef418bf` has the same tree as this head. The changed fixtures exercise true through object, dotted-flag and YAML input; formatting also passed locally. Local vet stopped before analysis because the pinned dependencies were unavailable, so no independent local loopback matrix or live API pass is claimed.

Approved.

saioai (approved):
the optional flag works in stable and beta, preserves false and omitted values, and keeps the other request options intact. independent checks and CI pass.

PR #201, opened 2026-09-18
feat(api): add webhook endpoint management

jbeckwith-oai:
## Summary

Adds commands for webhook endpoint management and event-type discovery.

## Changes

- Create, list, retrieve, update, and delete webhook endpoints
- Rotate signing secrets and send sample events
- List the webhook event types available to the project


markstuart-oai (approved):
Reviewed the webhook endpoint and event-type CLI commands at this commit. Create, retrieve, update, list, delete, rotate-secret and test map to the pinned Go SDK methods and schema; the list command uses the existing cursor iterator, and create/rotate preserve the signing secret returned only on those responses. The 1,230 added lines are chiefly generated OpenAPI schema, commands and tests, with a small command-registry change. I found no consequential issue.

Validation: two independent source-review passes, local `go vet ./...` with a writable module cache, `gofmt` and `git diff --check` passed; the PR build, lint, test, CodeQL and Castiron checks are green. I did not run a live API call or the local mock-server integration suite; the PR CI covers the full suite.

PR #208, opened 2026-09-20
fix(cmd): split PAGER into program and arguments

feiiiiii5:
## Summary

`$PAGER` is a command line in practice, but both pager implementations in `pkg/cmd/cmdutil*.go` resolve the whole value as a single executable name. With the very common `PAGER="less -R"`, any `... list` command whose output is longer than the terminal fails instead of paging.

## Problem

`streamToPagerWithPipe` and the Unix `openSocketPairPager` each did:

```go
pagerProgram := os.Getenv("PAGER")
if pagerProgram == "" {
	pagerProgram = "less"
}
pagerPath, err := exec.LookPath(pagerProgram)
...
cmd := exec.Command(pagerProgram)                       // "less -R" as one argv[0]
pid, err := syscall.ForkExec(pagerPath, []string{pagerProgram}, procAttr)
```

`git`, `gh` and `man` all document `PAGER` as `program [args...]`, and those tools are what most users set it beside, so `PAGER="less -R"` is the configuration this CLI meets in the wild. Both call sites then fail with:

```
exec: "less -R": executable file not found in $PATH
```

`streamOutput` has one production caller, `ShowJSONIterator` (`pkg/cmd/cmdutil.go:575`), which 68 generated list command sites return through, so the failure is not confined to one command. `streamOutputOSSpecific` does fall back to `streamToPagerWithPipe` when the socket pager fails, but the fallback resolves `PAGER` the same way, so both paths end in the same error.

## Fix

One `pagerCommand()` helper, used by both paths:

- trims the value, falls back to `less` when it is empty;
- splits the trimmed value with `strings.Fields` and returns argv when the first word resolves via `exec.LookPath`;
- if the first word does **not** resolve, keeps the whole value as one element, so a program path containing spaces (which does resolve today) keeps resolving.

The socket pager now passes that argv to `syscall.ForkExec`, the pipe pager to `exec.Command`.

**No shell is involved.** The value is split into argv words only; nothing is ever passed to `sh -c`, so `docs/architecture/security-model.md:88-91` (a local operator already controls `PAGER`) is unaffected in either direction: this neither adds an execution capability nor relies on one.

## Tests

`pkg/cmd/cmdutil_pager_unix_test.go` (new, `//go:build !windows` like the existing `cmdutil_output_errors_unix_test.go`):

- `TestPagerCommand` — 8 cases: unset, blank, bare program, bare program with surrounding whitespace, program plus two arguments, trimmed value with an argument, a path containing spaces, and an unresolvable first word.
- `TestStreamToPagerPassesPagerArguments` — drives **both** pager implementations through the same `paths` map already used by `TestStreamOutputErrorOrigins`, with a `#!/bin/sh` fixture that records `$1` and echoes stdin. `PAGER="<fixture> --no-init"`, and the test asserts the fixture received `--no-init` and the streamed bytes. The socket subtest asserts `w.Name() == "parent-socket"` so it cannot silently be covered by the pipe path.

Red on the base commit `0169bff` (test file copied into a scratch worktree of `0169bff`):

```
--- FAIL: TestStreamToPagerPassesPagerArguments/pipe
    Received unexpected error:
    exec: ".../001/pager --no-init": stat .../001/pager --no-init: no such file or directory
--- FAIL: TestStreamToPagerPassesPagerArguments/socket
    Received unexpected error:
    exec: ".../001/pager --no-init": stat .../001/pager --no-init: no such file or directory
```

Green on this branch, both subtests.

## Validation

cwd `/Users/fei/Desktop/开源/.runtime/2026-09-20/oa/src`, base `0169bff`, go1.25.0 darwin/arm64, network proxies unset, `-count=1`:

| Command | Result |
| --- | --- |
| `go test ./pkg/cmd -run 'TestPagerCommand\|TestStreamToPagerPassesPagerArguments' -v` | PASS (10 subtests) |
| `go test ./pkg/cmd -race -run 'TestPagerCommand\|TestStreamToPagerPassesPagerArguments\|TestStreamOutput\|TestShowJSON'` | ok |
| `go test ./internal/...` | ok, 0 failures |
| `go test ./pkg/cmd` | 284 failures — **identical set on base and on this branch** (diffed by test name: `new failures on branch: []`) |
| `./scripts/lint` | ok |
| `go vet ./...` | ok |
| `gofmt -l` on the three files | no output |
| `GOOS=windows GOARCH=amd64 go build ./...` and `go test -c ./pkg/cmd` | ok |
| `GOOS=linux GOARCH=amd64 go test -c ./pkg/cmd` | ok |
| `go mod verify` / `go mod tidy -diff` | all modules verified / no diff |

Not run, and why:

- `./scripts/test` (full `go test ./...`) needs the reviewed Steady mock server on `127.0.0.1:4010`, which downloads a pinned Deno and fetches the pinned fork. The 284 `pkg/cmd` failures are all the same precondition error, `Mock server is not running on localhost:4010`, plus one pre-existing failure I confirmed reproduces unchanged on `0169bff`: `TestFilesCreateCLICancelClosesStalledFIFO` → `cancellation must close the owned FIFO reader`. CI covers the mock-server suite.
- A real interactive TTY session. `streamOutput` only reaches a pager when `stdout` is a terminal, so under `go test` it short-circuits to `streamToStdout`; the new tests therefore call `streamToPagerWithPipe` / `streamOutputOSSpecific` directly, which is how the existing `cmdutil_output_errors_unix_test.go` tests reach them too.

Both touched files are wholly handwritten (no `Code generated` header), so nothing in the generated baseline changes; the Castiron custom-code report will post its own view of the diff.

AI-assisted: an AI coding agent produced this change under the account owner's standing instruction for this repo, ran every command above and read its output. No human reviewed the diff before it was opened.


saioai (changes requested):
check whether the complete PAGER value resolves before splitting it, add a regression test where both executable paths exist.

feiiiiii5:
Good catch. I fixed this in `81c0999`: `pagerCommand` now checks the complete `PAGER` value with `exec.LookPath` before splitting on whitespace, so an executable path containing spaces wins even when its prefix is also executable. I added a regression case with both paths present.

The same test fails on the previous head and passes here. I also ran `go test ./pkg/custom/...`, `go test ./... -run '^$'`, `go mod verify`, and `./scripts/lint`.

saioai:
[saibot]

hold for review. PAGER="less -R" now pages long output instead of failing, while executable paths with spaces still work. This small, dependency-free fix is justified by the head/base terminal reproduction. No blocking or substantive issues found. Hold for hosted checks and SDK owner review; Sai's earlier requested fix is addressed, but his changes-requested review remains active.

local checks: 5 passed; 1 base comparison failed; 1 check could not start.

limits: Linux/amd64 Go 1.27.0 only; no native macOS/Windows or Go 1.26 runtime verification. Windows custom-package compilation only. No full mock-backed scripts/test, live API calls, or manual real-less session. Runtime suite skipped six terminal-dependent cases…

github checks: 3 workflows awaiting approval.

approve workflows: [CI](https://github.com/openai/openai-cli/actions/runs/36026840147), [CodeQL](https://github.com/openai/openai-cli/actions/runs/36026840090), [Castiron custom code](https://github.com/openai/openai-cli/actions/runs/36026840069).

reviewed commit: `1c7fee46e64f`. base: `9d9076315350`.

feiiiiii5:
@saioai Done in 81c0999: `pagerCommand` checks the complete PAGER value with `exec.LookPath` before splitting it, and the regression case creates both the executable path and its executable prefix. The change remains on the current head `ff11fcf`.

PR #212, opened 2026-09-21
fix(cli): preserve complete plain-text audio responses

markstuart-oai:
## Summary

Preserves complete text, SRT, and VTT responses from audio transcription and translation commands.

## Changes

- Uses response content type to preserve plaintext even when the transcript resembles JSON.
- Keeps complete transcript and subtitle bodies, including Unicode and line breaks, through the existing output modes.


markstuart-oai:
@codex review

dpiet-oai (approved):

PR #216, opened 2026-09-21
fix(cmd): expand file references in nullable string flags

dipeshbabu:
Nullable string flags such as `responses create --instructions @file` currently send the reference literally. Expand nonnil `*string` values through the existing file-input path while preserving null, unset values, escapes, and untrusted-stdin provenance.

The same defect was covered by #168, which was closed without merging and remains present on current main. This change is limited to the handwritten helper and regressions using synthetic localhost requests.

Local validation on this isolated branch: CLI build, compilation of all test packages, nullable-input and untrusted-stdin regressions, `go vet ./pkg/cmd`, and the unchanged main custom-code budget checker (571/1000). The combined audit branch also passed the full mock-backed test suite and snapshot builds. Hosted CI is pending; SDK CODEOWNER review is required.


saioai (approved):
good, --instructions @notes.txt sent the filename literally, now it sends the file’s contents while preserving null values.

PR #219, opened 2026-09-21
fix(jsonview): propagate lazy-loading failures

dipeshbabu:
A failed lazy page currently leaves the explorer showing a truncated result and can return success. Save the load error, stop the explorer, and return it together with any terminal-program error. Normal iterator exhaustion still succeeds.

Local validation on this isolated branch: CLI build, compilation of all test packages, the complete `internal/jsonview` suite with the race detector, package vet, and the unchanged main custom-code budget checker (571/1000). Regressions cover root and nested views, normal exhaustion, and error propagation through the Bubble Tea event loop. The combined audit branch also passed the full mock-backed suite. Hosted CI and SDK CODEOWNER review remain pending.


saioai (approved):
looks good, this prevents failed requests from looking like complete results.

PR #220, opened 2026-09-21
fix(output): honor max-items in interactive lists

dipeshbabu:
Interactive list output ignores `--max-items` while preloading and fetching later rows. Wrap the iterator with the requested count so neither path reads beyond the limit; zero items returns without consuming input, and an unlimited iterator keeps its existing behavior.

Local validation on this isolated branch: CLI build, compilation of all test packages, iterator-limit regressions, `go vet ./pkg/cmd`, and the unchanged main custom-code budget checker (571/1000). Tests count upstream reads and verify error propagation. The combined audit branch also passed the full mock-backed suite. Hosted CI and SDK CODEOWNER review remain pending.


saioai (approved):
looks good, this makes the zero-item case return without opening the explorer or writing output, the regression tests pass.

PR #221, opened 2026-09-21
fix(completion): resolve command aliases while traversing commands

dipeshbabu:
Completion suggests command aliases but does not recognize them when selecting the command whose flags should be completed. Match visible commands against all their names so `cfg:get --sel` completes the same flag as `config:get --sel`.

Local validation on this isolated branch: CLI build, compilation of all test packages, the complete `internal/autocomplete` suite, package vet, and the unchanged main custom-code budget checker (571/1000). The regression checks canonical and alias command names. The combined audit branch also passed the full mock-backed suite. Hosted CI is pending.


saioai (approved):
looks good, autocomplete now recognizes command aliases too, the regression test passes.

